PR #6, opened 2018-08-15
sysinfo: armhf/arm64 compatibility

jesset:
- ALSAVER/SOXVER
- CPU version
- Board revision

works on both armhf/arm64


moodeaudio:
Hi @jesset,

I've added this commit to upcoming moOde 4.3 release.

-Tim

PR #7, opened 2018-09-04
browser graidview and Some optimization

lazybat:
browser graidview and Some optimization

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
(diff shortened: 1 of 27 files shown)


monkeypox:
Hello @lazybat,

There is significant code drift between the current release of moode (4.2) and the development version (4.3g) which makes a direct merging problematic. I realize that makes contributing difficult. It would be helpful if you could post screen shots of the non-bugfix changes that you have submitted and would like to see merged into moode so we can think about how best to add them. As Tim mentioned we’ve already done an album covers view but that doesn’t mean we can’t use any of your work in that and other areas.

Ideally we can all work together to make moode better.

moodeaudio:
Hi @lazybat (and @jesset),

When moOde 4.3 is released, please view the contributors page via Menu, Configure, System then Menu, About, View contributors.

Your Git userid's will be listed along with a brief description of your idea and code contributions in accordance with Open Source community practices. Also there will be attributions in the source code headers where your code contribs are used.  :-)

-Tim

moodeaudio:
Album Cover panel and associated Thumbnail generator are part of moOde 4.3.

PR #13, opened 2018-12-24
Extend zero duration handling for floats

ohinckel:
In some cases the duration will be set to 0.00 (float) value which results
in a infinity float for the duration fields.

The check for the zero duration will be extended to not only check if the
field is present, instead it will also check if the value is zero.

In some cases, especially when playing a stream, the duration was set to
0.0 which was producing JavaScript errors and the UI was not usable anymore.

### DIFF
--- a/www/inc/playerlib.php
+++ b/www/inc/playerlib.php
@@ -707,7 +707,7 @@ function parseStatus($resp) {
 			$array['time'] = '0';
 		}
 		// radio, upnp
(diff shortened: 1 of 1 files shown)


moodeaudio:
I've integrated this PR into moOde 4.5 code base and upcoming moOde 4.4 update.

Thanks for the contrib :-)

ohinckel:
Great, thanks too.

PR #14, opened 2018-12-24
Handle JSON encoding errors in engine-mpd and show then in UI

ohinckel:
The engine-mpd.php will generate a error structure in case an error happens
while encoding/generating the JSON response data (e.g. due to invalid
values which can not be encoded, like NaN or Infinity values).

In the UI this case of error is now checked and handled to show an error
notify to the user. This includes the JSON error message and JSON error code
for easier debugging/fixing.

In some cases it could happen that the generated data which should be converted
to JSON contains some invalid values (as mentioned above). Usually this should not
happen, but if the user will see the error.

### DIFF
--- a/www/engine-mpd.php
+++ b/www/engine-mpd.php
@@ -99,6 +99,11 @@ debugLog('engine-mpd: Metadata returned to client: Size=(' . sizeof($current) . 
 // TEST I don't think this is needed
 //header('Access-Control-Allow-Origin: *');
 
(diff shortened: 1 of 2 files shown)


moodeaudio:
I've integrated this PR into moOde 4.5 code base and upcoming moOde 4.4 update.

Thanks for the contrib :-)

ohinckel:
Thanks for this integration too.

Issue #2, opened 2018-04-08
"RPC Error: Unable to receive"


webodan:
I've just installed moode on a pi1, and I'm unable to mount my NAS. Neither SMB nor NFS will work, and this NAS works flawlessly with several other machines (including volumio). in SMB I get "RPC Error: Unable to receive", and in NFS I get "Mount error: (mount error(13): Permission denied". I don't know what else to do. Could someone help me?

Thanks in advance.

PR #16, opened 2019-02-11
Group albums by album_artist

chris-rudmin:
- Return flat list from playerlib.php
- Add artist, album_artist, and lastModifiedTime to metadata
- Create albums based on album_artist, fallback to artist.
- Remove config and code for compilation rollup.
- ~~Sort Album thumbnail view by last_modified date~~


www/js/playerlib.js
```diff
@@ -2194,26 +2144,20 @@ var renderSongs = function(albumPos) {
...
-  	// Library panel cover art and metadata
+  // Library panel cover art and metadata
 	if (allAlbums.length == 1 || LIB.filters.albums != '' || typeof(albumPos) !== 'undefined') {
 		$('#lib-coverart-img').html('<a href="#notarget" data-toggle="context" data-target="#context-menu-lib-all">' + 
 			'<img class="lib-coverart" src="' + makeCoverUrl(allSongs[0].file) + '" ' + 'alt="Cover art not found"' + '></a>');
 		$('#lib-albumname').html(allSongs[0].album);
 
-		// r44d for compilation album when using Artist instead of AlbumArtist list ordering
-		if (typeof(albumPos) !== 'undefined') {
-			if (albumPos = UI.libPos[0]) {
```
chris-rudmin: @moodeaudio This is super confusing doing an assignment in an if statement, and I think there might a bug here. Should it be `if (albumPos === UI.libPos[0]) {`? I think my new code follows the same behaviour as the old.

www/inc/playerlib.php
```diff
@@ -409,32 +400,25 @@ function genLibrary($flat) {
 function genLibraryUTF8Rep($flat) {
 	$lib = array();
 
-	// use Artist or AlbumAetist for the Artist column
-	$libartist = $_SESSION['libartistcol'];
```
chris-rudmin: @moodeaudio I removed `libartistcol` here and in the js, but I'm not sure if there is more cleanup to do (db migration or something)

www/js/playerlib.js
```diff
@@ -1770,77 +1770,99 @@ function loadLibrary() {
...
+  };
+
+  var reduceAlbums = function(acc, track) {
+  	(acc[track.album] = acc[track.album] || []).push(track);
+  	return acc;
+  };
+
+  var genreFilter = LIB.filters.genres;
+  if (genreFilter.length) {
+  	allSongs = allSongs.filter(function(track) {
```
chris-rudmin: Doing filtering here on a flat structure will give us more control should we want different sort/filtering views.

www/inc/playerlib.php
```diff
@@ -364,37 +364,28 @@ function genFlatList($sock) {
...
-		$artist = $flatData[$libartist] ? $flatData[$libartist] : ($flatData['Artist'] ? $flatData['Artist'] : 'Unknown');
-		$album = $flatData['Album'] ? $flatData['Album'] : 'Unknown';
-		//$album = $flatData['Album'] ? $flatData['Date'] . ' - ' . $flatData['Album'] : 'Unknown'; // r44g add year
-		//$album = $flatData['AlbumSort'] ? $flatData['AlbumSort'] : ($flatData['Album'] ? $flatData['Album'] :'Unknown'); // albumsort tag if present
-
-		if (!$lib[$genre]) {$lib[$genre] = array();}
-		if (!$lib[$genre][$artist]) {$lib[$genre][$artist] = array();}
-        if (!$lib[$genre][$artist][$album]) {$lib[$genre][$artist][$album] = array();}
 
 		$songData = array(
```
chris-rudmin: @moodeaudio MPD docs says [we should not manage a copy of the db](https://www.musicpd.org/doc/html/protocol.html) (and i agree). This is what we are doing here. An improvement could be made to send sort filter queries directly to the db as we need.

moodeaudio:
Hi Chris,

Thx for the PR. I'll have a look this week.

-Tim

www/js/playerlib.js
```diff
@@ -1854,7 +1876,7 @@ function filterLib() {
 			allArtists.sort(function(a, b) {
 				return collator.compare(removeArticles(a), removeArticles(b));
 			});
-			allAlbumsTmp.sort(function(a, b) {
+			allAlbums.sort(function(a, b) {
 				return collator.compare(removeArticles(a['album']), removeArticles(b['album']));
```
chris-rudmin: @moodeaudio I noticed that `removeArticles` only removes english articles. Wouldn't it make sense to use `album_sort` or `artist_sort` tags instead?

moodeaudio: Are those tags normally populated in tracks or by ripping software?

chris-rudmin: Not as far as I know. However it would give the user back control of how they want their library sorted. And these tags are created for specifically this use-case. It would avoid an infinity of english language centric bugs being reported. (eg: how come we sort with `Beatles` instead of `The Beatles` but not `Tigre` instead of `Le Tigre`)

chris-rudmin: IMO, If those tags are populated, then we should use them. Perhaps we could fallback to removing english only articles.  This also opens the possibility for users to sort artists `lastname, firstname` or other such OCD schemes.

moodeaudio: It makes sense to use them if populated. We just need a way to let the client know whether or not to use removeArticles().

moodeaudio:
@chris-rudmin,

If you join the moOde Forum and send me your userid I'll add you to the test team section where links to the latest test image and sources are posted.

-Tim

chris-rudmin:
@moodeaudio @moode-player/contributors I've setup a pi for testing, and I've given this code some more thorough testing/bugfixes. Tested on Chrome, iOS, Firefox, Safari. Behvaiour should be the same as before (except that albums are properly grouped by album artist)

If artist and album artist are different for a specific album, the songs will show up under both now (in the genre/artist filter view)

moodeaudio:
Hi Chris,

I'm planning to start working with this PR after moOde 5 ships :-)

-Tim

chris-rudmin:
@moodeaudio Merging this PR with the v5.0 changes was super painful. I will need to do some more validations.

moodeaudio:
Understand. I'm open to a process that can get our nice project into full Git development mode. Currently we only use Git for batch updating that final release.

www/js/playerlib.js
moodeaudio: moOde 5 includes a setting on the Library Options screen to allow user to define the list of articles to be ignored or if user blanks out the list then no articles would be ignored. 

I also included logic in function genLibrary($flat) to substitute ArtistSort and AlbumSort tag values for the default tag values Artist/AlbumArtist and Album.  Apparently this is bad. Users posted in moOde Forum that only part of their collection had the sort tags populated and this really messed up the sorts in the Library views. Partial population occurs when users have used various tagging utils on their collection. Some utils populate the tags (Picard) and other don't. 

Other comments were that the sort tags should only be used for sorting and that the default tags should be used for display. 

That makes sense and this could certainly be done but it would only work if the entire collection had the Sort tags populated. If the collection had partially populated tags it would turn out bad.

Id like to use these tags but I think it would require a setting with a help that explained that ALL the sort tags have to be populated for this to work.

-Tim

monkeypox:
I haven’t looked at the commit but if it doesn’t work this way already, maybe change the library option to have sort options as Artist, Album Artist, Artist Sort and Album Artist Sort and ignore the sort tags when appropriate.

www/js/playerlib.js
chris-rudmin: @moodeaudio It's not quite clear why partially tagged collection wouldn't sort correctly. `allAlbums.sort((a,b) => (a.sortArtist ||  a.artist).localeCompare(b.sortArtist || b.artist))`

chris-rudmin: I agree that the sortArtist and the sortAlbum tags should only be used for sorting.  The displayed name should always be artist or album.

chris-rudmin:
> Understand. I'm open to a process that can get our nice project into full Git development mode. Currently we only use Git for batch updating that final release.

Do you already have an idea of what workflow you would like?  If you want, I would be happy to share my thoughts and experiences.

TheOldPresbyope:
It seems an impossible goal to have moOde make up for mixed, missing, and bad metadata in users' collections. It's on the user to regularize their collection one way or the other.

So, I could imagine a configuration option which allows me to choose

Case 1: sorting as needed on album / artist / albumartist values
Case 2: sorting as needed on album_sort / artist_sort / albumartist_sort values; falling back on a regular value only if a _sort value is missing.

In either case I would display based only on album / artist / albumartist values.

I for one would always use Case 2 if it were available. Modern ripping programs do a good job filling in the values. (But I've ripped CDs over a decade. I have to go off and write a script to scan my collection to see how of the early rips had the _sort values filled in!)

As an aside, we are using the word "tag" colloquially. In most tagging schemes, the word "tag" refers to the entire metadata block.

www/js/playerlib.js
moodeaudio: When only some of the sort tags are populated we can end up with something like below

Artist = Tim Curtis | ArtistSort = <does not exist>' | Album = Blue Sky
Artist = Tim Curtis | ArtistSort = Curtis, Tim | Album = Red River

Artist = exists(ArtistSort) ? ArtistSort : Artist

In this case the artist Tim Curtis will end up appearing twice in the artist list, once under the T's and again under the C's

chris-rudmin: @moodeaudio This seems correct given inconsistent tagging.

chris-rudmin:
@moodeaudio These changes works with Moode 5.0, but with 5.1 crashes.  Could you release the source for 5.1 so I can do another code merge?

www/js/playerlib.js
```diff
@@ -1664,195 +1679,119 @@ function loadLibrary() {
...
 	}
-	// only one album in list
-	else if (allAlbumsTmp.length == 1) {
-		var objRegularAlbum = {'album': allAlbumsTmp[0].album, 'artist': allAlbumsTmp[0].artist, 'compilation': '0', 'imgurl': allAlbumsTmp[0].imgurl}; // store the one and only album
-		allAlbums.push(objRegularAlbum);
+
+	// sort Genres, Artists, Albums
+	allGenres.sort();
+	allAlbumCovers.sort(function(a, b) {
+		return b.last_modified - a.last_modified;
```
chris-rudmin: @moodeaudio We don't need to force this change on everyone. This is just my personal preference for sorting (newest first)

chris-rudmin:
> @moodeaudio These changes works with Moode 5.0, but with 5.1 crashes. Could you release the source for 5.1 so I can do another code merge?

Thanks!  Working now with 5.1.

chris-rudmin:
@moodeaudio If there is there anything stopping you from pulling in this change, I would be happy to address it.

moodeaudio:
Lack of time. The dust from the moOde 5 release has not yet settled. My primary responsibility to users is to fix major bugs that they report for this new release. I was hoping that the 5.2 release would be the "stability release" but unfortunately just today there is a report that one of the main network features, "AP Mode"  is not working :-(

I'm able to repro this and so I need to devote my cycles to debugging it and providing a fix. I'm hoping it's a config issue and not some sort of Kernel or Raspbian firmware regression. Ugh.

pccasto:
I changed the base of this pull request from master to develop. Since they are both the same right now, this seems like the best time to make the change.

chris-rudmin:
@moodeaudio Do you have any plans to merge this? If not, then I will stop maintaining this branch and fixing the merge conflicts.

moodeaudio:
I'm planning to test it this week.

moodeaudio:
@chris-rudmin,

No need to do any fixing. I'm going to manually merge and test your enhancements :-)

moodeaudio:
I've Integrated the changes and initial tests are looking good :-) The libcache.json file is ~30% larger but it doesn't seem to affect Library load time, at least for my 4000 track collection. Navigating around the Library is fast and responsive.

Comments and Questions:

1. There are a couple of trivial mods I need to make to accommodate the new moOde 6 options for Artist and Album list ordering. User can specify Artist, ArtistSort, AlbumArtist or AlbumArtistSort tag for the Artist list, and Album or AlbumSort for the Album list.

2. There is a virtual tag named actual_artist in the old code thats used in place of the artist tag for display purposes for example it enables the actual artist to be displayed for compilation albums. I think this tag needs to be added back. It looks like another trivial mod.

3. I'm still unclear how automatic compilation rollup happens without the original routine.??

4. I think we need to test against a large Library.

-Tim

chris-rudmin:
For number 2, the artist can be found under ‘artist’ . I’m not sure why the
data would need to be duplicated.

chris-rudmin:
For number 3 albums are grouped by the ids, album@album_artist or
album@artist if album artist is not defined.

moodeaudio:
For 2 we need to retain the actual_artist because we are allowing the user to choose which tag is used to populate 'artist'. Same for 'album'. It also depends on how user has decided to deal with compilation albums in their collection. A common approach is to set AlbumArtist = "Various Artists" for compilation albums, and Artist for regular albums, then specify that the AlbumArtist tag is used to order the artists list. This approach also eliminates the need to do any type of automatic compilation rollup.

```
	$lib_artist = $_SESSION['library_artist_col']; // Artist, ArtistSort, AlbumArtist or AlbumArtistSort
	$lib_album = $_SESSION['library_album_col']; // Album or AlbumSort

	foreach ($flat as $flatData) {
		$songData = array(
			'file' => $flatData['file'],
			'tracknum' => ($flatData['Track'] ? $flatData['Track'] : ''),
			'title' => ($flatData['Title'] ? $flatData['Title'] : 'Unknown Title'),
			'disc' => ($flatData['Disc'] ? $flatData['Disc'] : '1'),
			'artist' => ($flatData[$lib_artist] ? $flatData[$lib_artist] : ($flatData['Artist'] ? $flatData['Artist'] : 'Unknown')),
			//'album_artist' => $flatData['AlbumArtist'],
			'actual_artist' => ($flatData['Artist'] ? $flatData['Artist'] : 'Unknown'),
			'composer' => ($flatData['Composer'] ? $flatData['Composer'] : 'Composer tag missing'),
			'year' => $flatData['Date'],
			'time' => $flatData['Time'],
			'album' => ($flatData[$lib_album] ? $flatData[$lib_album] : ($flatData['Album'] ? $flatData['Album'] : 'Unknown')),
			'genre' => ($flatData['Genre'] ? $flatData['Genre'] : 'Unknown'),
			'time_mmss' => songTime($flatData['Time']),
			'last_modified' => $flatData['Last-Modified']
		);
``

Heres a few pics that illustrate the situation.

New code
![Screen Shot 2019-08-02 at 5 36 19 PM](https://user-images.githubusercontent.com/13685583/62400144-33a36900-b54c-11e9-9eeb-bd8e05017595.png)

Old code
![Screen Shot 2019-08-02 at 5 36 34 PM](https://user-images.githubusercontent.com/13685583/62400150-38681d00-b54c-11e9-8dd0-975f6de7096f.png)

**Comment by chris-rudmin** [comment]
I would leave artist as artist and for the usecase you want to support, the field would be display_artist or something like that.

**Comment by monkeypox** [comment]
On my phone so I can’t check the PR on live moode but the reason there needed to be actual_artist is that artist is filled with the album artist tag (if it exists) in the php function if album artist is the chosen sort method. That makes things easier when constructing the lists, etc. since there’s only one artist code path. The downside is you need to reload the library if you change the sort type.

Assuming everything behaves as before with Chris’s code wrt artist / album artist sort behavior and there isn’t the afore-mentioned mingling of data then it’d be easy to just use artist instead of actual_artist in the function that generates the track list as they’re the same data.

I’ll try to get moode 6 up this weekend and apply the patch to see how it works on my medium sized (24k tracks) library.

**Comment by chris-rudmin** [comment]
Yeah my concern is that this PR uses artist field to do the grouping. To
avoid duplicate data in the getlibrary API could we do the display artist
logic in js?

**Comment by moodeaudio** [comment]
Right. the 'artist' and 'album' fields returned from the PHP code are only used for group/sort in JS thus there needs to be 'display_artist' and 'display_album' fields containing the actual artist and album names.

After modifications I have the Library working same with new and old code. The new code results in ~40% larger libcache.json file cos of its structure and additional fields which means more data over the wire during Library load and more Browser memory to hold the data. 

Old
```
pi@rp2:~ $ ls -lh /var/local/www/libcache.json 
-rwxrwxrwx 1 root root 991K Aug  2 21:59 /var/local/www/libcache.json
```
New
```
pi@rp3:~ $ ls -lh /var/local/www/libcache.json
-rwxrwxrwx 1 root root 1.6M Aug  3 08:00 /var/local/www/libcache.json
```

@monkeypox - I'll post the two files later today so you can try out on your 20K track collection. Note though that we have a lot of users with 2X or 3X that number of tracks!

monkeypox:
I’ll try to get it set up tonight.

If we need display fields and size is an issue then maybe we can optimize by only storing data in the display fields if they’re different from the sorting fields? That’d add a bit more work in javascript to return the right value but shouldn’t be a big deal.

moodeaudio:
I've discovered that a bit more work is needed to the current Lib code to accommodate the new sort options and display fields so just hang tight for a few days.

We have to plan around the display fields being different from the sort fields. If size over the wire is a concern then maybe compression can be used.

chris-rudmin:
If you want to optimize your API I would recommend to split it up to drive
different parts of the ui. One for album, one for genres, artists, one for
album tracks.  It’s not really necessary to load all tracks with all
metadata up front.

moodeaudio:
I separated out the lib code from playerlib.js into scripts-library.js and scripts-library-new.js and was able to A/B the current and new code under more scenarios.

- The new code does not handle the case where Artist is used to order the artist list and an album contains tracks with different artists i.e., a compilation album. The album appears multiple times, once for each artist. This is the scenario where user collection does not have the AlbumArtist tag populated and thus has to use Artist. This is what drives the need for an auto-compilation-rollup routine.

- The current and new code are not set up to handle the new Artist and Album sorting options that I added to Lib Options config. Using display_ fields works to a point then becomes very messy.

Bottom line is that moOde 6 might ship with the existing Library and no additional sort options just for stability until we can work out the new Lib.

chris-rudmin:
I’m of the opinion that albums of the same name, but with different artists
and no album artist tag should NOT be grouped together.

moodeaudio:
That would leave some users with no way to correctly display their compilation albums.

chris-rudmin:
They should tag their compilation albums with album artist. Grouping albums
indiscriminately by album title will incorrectly group albums with the same
title, “greatest hits” is a common album title that comes to mind that
should not be grouped.

How do other media players handle albums of the same title with different
artists and no album artist tag?

moodeaudio:
User can enter an Exclusion list in Library options. It defaults to "greatest hits". The exclusion list only applies if Compilation Rollup is set to Yes.

The only other player I've used in recent years is iTunes. Every compilation CD I've ripped with iTunes was imported correctly and displayed as a compilation Album. I assume the compilation indicator comes from the metadata lookup database thats used during the rip. There is also a "compilation" tag thats set in either the tracks or the iTunes Library DB, I can't recall which one. It's unique to iTunes. The AlbumArtist tag is often not populated by iTunes CD ripper.

When MPD indexes one of these albums it has no way of detecting that it's a compilation. This is because MPD does not support a compilation tag, and this is because there is no id3 standard for it.

This is why Compilation Rollup option was added to moOde way back when. It compensates for users and rippers not setting the AlbumArtist tag to "Various Artists".

moodeaudio:
Actually, It might be worth asking on the Forum if anyone still uses Compilation Rollup. Maybe it's outlived its usefulness.

moodeaudio:
I posted a link to test image #2 in the Test Team Section of the Forum. It includes the new Lib code :-)

monkeypox:
Another data point: I turned compilation roll up off (and fixed tags that needed fixing) and think that’s probably the right approach for most people, especially those with larger collections where album titles are more likely to conflict. I always used album artist tags for compilation albums though so it wasn’t as much of a struggle as it might be for some.

moodeaudio:
Good info. I wonder why there was never a standard id3 tag for compilations...

moodeaudio:
@monkeypox, have you tested the #2 image yet on your collection?

chris-rudmin:
Just FYI the reason I created this pull request was to fix an issue where
albums were duplicated if tracks contained different genres. There is a
forum thread describing the issue.

moodeaudio:
I'd forgotten about that. i'll consider it an additional bonus :-)

moodeaudio:
@chris-rudmin - This PR has been integrated into moOde 6 develop branch.

chris-rudmin:
@moodeaudio I think it would be nicer for the contributing developers if you merged their commits so that they get recognition for their code changes. It is also easier for you to track the changes as you would have all the commit history and commit messages.

For code reviews, it is acceptable that you can ask for changes if it does not meet your standards, and then you merge the commits when you are satisfied with it.

I don't think it is acceptable to steal peoples code changes, and pass them off as your own (which is how this looks).

moodeaudio:
Thats odd. I manually merged your PR so you wouldn't have to keep fixing it to the other updates I was committing. I thought I was doing you a favor. I also had to pull out the old and new library code into separate files to be able to A/B test and verify that your new Library routines generated the same Library in the UI as the old routines. This was a lot of work btw, and work that you did not have to do. In any case your contribution is already attributed in the code, just search for chris-rudmin.
https://github.com/moode-player/moode/blob/develop/www/CONTRIBS.html
https://github.com/moode-player/moode/blob/develop/www/inc/playerlib.php
https://github.com/moode-player/moode/blob/develop/www/js/scripts-library.js

If you want different wording, just let me know. I suppose another option would be for you to modify the files in some way or add your own contrib wording and then resubmit them as a new PR and I'll just merge that. 

-Tim

chris-rudmin:
@moodeaudio Thanks Tim, I am not specifically looking for recognition, but wanted to bring to your attention the norms (as I understand them) around contributing to open source projects.

If you are uncertain about merging commits using the command line, I would recommend using the github UI.  At the bottom of the pull request, there should be a big green button "Merge". Don't be afraid to use this :)

moodeaudio:
lol, the moOde project is anything but normal :-)

Formal Git processes are super solid but when it's 10X easier and faster for me to integrate and test a contrib manually I'll go that route. I already use Git Desktop and Atom editor. it's a really nice tool set.

PR #17, opened 2019-02-17
Cue action icon

pccasto:
A small bug-fix for the handling of icons for 'cue' playlists when navigating using the directory/file view.

This is a belt-and-suspenders fix, with a change to the js-generated html to include 'float:left' and also the addition of a CSS to render that explicit styling unneeded. This stays consistent with the current approach, but should allow for the removal of this, and other explicit "styles=", in the future.

Thanks for this player!


pccasto:
Closing this, because it was a PR against master. Have created a new one ( #29 ) against develop.

PR #19, opened 2019-02-18
Feature Addition: flac Embedded Cue handling

pccasto:
Hi Tim/@moodeaudio

Thanks again for MoOde.  This pull request is now seeming to take longer to describe than the code diff it explains. However, I figure it is better to attempt to cover all the bases. And even so, I'm open for questions, and discussion where I missed something...

**Summary**
Allow MoOde to treat flac files with embedded cue sheets as playlists, when handing them off to `mpd`, so that they are added with a by-track rendering, rather than as a single-track/complete-album basis.

**Description of Change**
The attached pull request allows for special handling of flac files with embedded cue sheets from the browse/folder view (the challenging part :-) ) and from the library & album views. This allows users to add flac files to the player, and have them fully expanded to track-based playlists, rather than single flac-file based playlists.

The key change is an extra layer of logic in playerlib.php to determine if the flac has an embedded cue sheet, and then set the mpd command to `load`, rather than `add`. Other changes ensure that a list of files is passed as an array, rather than just the top-level path, and that the items in the array are individually iterated through the add/load logic.

**Environment**
I use flac with both embedded and external cue sheets, stored in a directory structure artist/album.
Each of those directories contains `   file.flac    file.cue   folder.jpg`.
When multiple discs in album, structure is `   file(disc1..n).flac      file(disc1..n).cue     folder.jpg`

[I know I can simplify that some, but have needed this structure to deal with handling in various programs, including MoOde]

**MoOde's current behavior:**
When browsing this environment via directory structure, a folder can show with three entries for only two files (*N if folder has multiple disc files):

- A 'phantom entry' for the embedded flac from file.flac (if the flac has one), with a playlist icon, and the full path of the flac file 'NAS/mount/album/artist/file.flac'  displayed in the GUI.  This has some 'buggy'/counterintuitive consequences...
-- User can select a menu choice of deleting the playlist (never tried it... expect it would fail in a variety of ways, but why take chances...). 
-- User can select menu choice of adding/playing/etc. All of those choices result in the file.flac file being passed to mpd as a path/file.flac with the `add` command, so the entire file.flac is treated as single track.
-- User can drill down into the playlist and see the number of tracks, each with the original file.flac as the reference. No per track info is displayed. If I recall correctly, an add at this level adds the entire flac file as a single album.

- The file.flac (with embedded cue, or not) - treated as a song file with a standard single song-file/track.
-- User can select adding/playing/etc. All of those choices result in the file.flac file being passed to mpd as a path/file.flac with the `add` command, so the entire file.flac is treated as single track.

- The file.cue presented as a Cue sheet/playlist
-- User can select adding/playing/etc. All of those choices result in the file.cue being passed to mpd as a path/file.cue with the `load` command, so the referenced file.flac is loaded as separate tracks by mpd, and displayed as such.
-- This was/is my desired behavior for all, and the reason I have kept external file.cue files, and the reason I always used the browse/folder view.

- At the folder level
-- The add/play/etc passes the path to the folder, which is passed along to mpd as an `add`, so only files are brought in by mpd, and with the `add`, they are not expanded into tracks, even if they have an embedded cue sheet.

**When browsing via the library or album view:**
- Only the files that mpd lists as 'files' are displayed, so the handling is as the same as for the second item above (The file.flac as a song file with handling as a standard single song-file/track).
- When adding multiple song-files from the album/artist view, they are passed as an array of files, but lumped together as a chain of `add` commands, so they are not expanded into tracks, even if they have an embedded cue sheet.

**MoOde's behavior after the pull request**
When browsing via directory structure, a folder will show with two entries (the phantom entry for the embedded cuelist is no longer displayed).

- The file.flac as a song file icon with handling as playlist if it contains an embedded cue.
-- User can select adding/playing/etc. All of those choices result in the file.cue being passed to mpd as a path/file.flac with the `load` command (not `add`), if it has an embedded cue, so the file.flac is loaded as separate tracks by mpd, and displayed as such.

- The file.cue presented as a Cue sheet/playlist
-- [Unchanged] User can select adding/playing/etc. All of those choices result in the file.cue being passed to mpd as a path/file.cue with the `load` command, so the referenced file.flac is loaded as separate tracks by mpd, and displayed as such.

- At the folder level
-- The add/play/etc gets a list of mpd files (`File:`) from mpd, and passes the array downstream, rather than sending just the starting path. This allows for each file to be evaluated, and if the file is a flac with embedded cue sheet then mpd `load` handling is called, resulting in track expansion for each of the files.

**When browsing via the library or album view:**
- Only the files that mpd lists as 'files' are displayed, but now when the path to the file.flac is passed downstream, extra logic determines if the file has an embedded cue sheet, and if so calls mpd with `load`. This results in expansion to track listings.
- When adding multiple song-files from the album/artist view, they are passed as an array of files, but now that array is iterated on a per-file basis, with handling as per the point above.

**Considerations**
Depending on your thoughts on these, I can refactor the code/pull request accordingly.
- Would everyone want this handling? I would think so, if they went to the trouble of embedding cue sheets in their flac files... but if not, a user configuration could bypass the test `if ($ext == 'flac')` in `addToPL` in playlistlib.php.
- I added a function `listall` / `listAll` to use/mimic the mpd command of the same name. It works, howvever in retrospect this could use the existing `lsinfo` with additional parsing in playerlib.js to do the selection of element == 'file'.
- I accidentally included one of the changes for a bug submitted under a different pull request in this branch, but have reverted it here, to keep things isolated. This is independent of the other two pull requests.

**Potential Future Enhancements**
1.  [UI] User configuration to not display external cue sheet, in the browse view, if it corresponds to a flac with an embedded cue. With this change to file.flac handling, there isn't a reason to show/use the file.cue (unless a user does not have embedded cue sheets). Unless point 2 blow is implemented before point 3.
1.  [UI] Ability to drill down into external cue sheets, and select specific tracks.
1.  [UI] Ability to drill down into flac files with embedded cue sheets, and select specific tracks.
1. [UI] Menu item for 'add/play/etc as album' to allow users to send flacs with embedded cue sheets to mpd in their entirety without adding by-track. Not sure if users would want this, but it is a possible extension.
1. [Other formats] This logic may be extensible to wavepack / .wv. I don't use it, so haven't experimented there.
1. [Refactor] The call to `mpd listallinfo` does return enough info to determine if a flac file contains an embedded playlist (`genflatfile` in `playerlib.php` could extract/parse that rather than skip it). That info could be pushed up into the data-set, and then used when sending the various play/add calls. Or, at the very least, this data could support point one above - UI knows that a flac file has an embedded playlist, and can allow for different icon and/or menus.
1. [Refactor] Separate out the sort logic from renderBrowse, so that it can also be used when adding all files in a folder. Then update the folderadd/folderplay to return files in same order as they would be displayed.

Best regards,
Paul


moodeaudio:
Hi Paul,

Really nice PR :-)

I'm tight for time though and won't get to this till after moOde 5 ships. As i mentioned to @chris-rudmin, the author of the Album sort PR, our project only uses Git for archiving our final release sources and image. All the development, exchange of sources and testing is done locally. It's very inefficient but migrating development to Git always seems to get bumped down on the TODO list...

In the meantime, I can add you and @chris-rudmin to the test team section in the moOde Forum. This is where links to the latest in-process test release images and sources are posted.

You will need to join the Forum and send me your userid

-Tim

pccasto:
Hi Tim,

Thanks.  I figured this PR might take a bit of time to review, and validate. I have been a forum lurker / user of MoOde for a number of years, and only noticed on Friday that you had the code at GitHub (for over a year now...).  I was finally upgrading a system from 4.0b12 to 4.4 ...

Based on the way the code was committed (in batches to GitHub), I figured you had an outside process for managing the work. However seeing the release code in GitHub was what inspired me to do a dive into the code over the weekend, and submit the recent PRs.  So even though the development process is external, I'm glad the production code is accessible in this manner.

I have a few more PRs the wings. One of which is a bugfix that may be of interest in the short term. I'll try to get that in soon. The others are enhancements that can wait - particularly since they aren't even close to coded yet...

Cue sheet handling hasn't worked the way I've wanted it to for a long time with MoOde, but same can be said of several other programs :-). I can wait another release cycle for that to change. Particularly now that I've dug in to figure out why that is, and know how to patch it the way I want it to behave...

I've just registered on the forums as 'pcasto'. I read the PR from @chris-rudman, and have some ideas that I want to contribute after his work is merged, to minimize conflicts.

Best regards,
Paul

pccasto:
Need to shift this as a PR to the develop branch. I might have to create a new PR.

This also needs some conflicts resolved along the way.

pccasto:
I've changed the base for the PR, but now need to go back and correct any merge conflicts.

pccasto:
I went through and resolved conflicts, but need to do some testing! There was one place where it looks like there was a change in functionality -- some code that used to run on the browse panel may no longer run there. But this PR might have expected it to.

pccasto:
Closing in favor of PR #46. The other branches had fallen fairly far behind, and the code in the new PR will be easier to test.

PR #20, opened 2019-02-19
Versioned resource files

pccasto:
Hi Tim,

I had an issue after changing a system from 4.0b12 to 4.4 with a clean/new install.  Browser would not display pages until I cleared the browser cache.  After solving the problem by clearing the cache, I notice that someone in the forums had the same issue.

Attached fix provides **a** way of adding some versioning data to the .css and .js files, so that they will be treated as static, but updated with new releases.

I'm not a PHP person... so there might be better recipes for doing this...   The function code was added into playerlib.php. Not sure if that is the best spot for it, but that file is included at the top level, so was the most straight forward. (Is there another file for html helper code?). 

Also, the functions have hard-coded version strings that have to be maintained individually (uggg!). Best approach would be to pull version from config DB.  Rather than wait until I figure out how to pull that data, I thought I'd get this PR submitted, so it could be considered for the upcoming release. 

It may need some rework to be properly integrated. But figure I would through this out for consideration since the bug it fixes is related to consequences of upgrade.

Change, as coded, does produce pages that are as expected. (Unchanged, except for the addition of a version tag to the .css and .js resources in header.php and footer.php respectively). Once in place, a browser that was previously choking on opening the site (because it still had resource files cached) was able to cleanly load the site.

~paul

### DIFF
--- a/var/local/www/header.php
+++ b/var/local/www/header.php
@@ -45,18 +45,19 @@
 
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0 user-scalable=no">
-	
-	<link href="css/bootstrap.min.css" rel="stylesheet">
-	<link href="cssw/bootstrap-select.css" rel="stylesheet">
-	<link href="cssw/flat-ui.css" rel="stylesheet">
-	<link href="css/fontawesome-moode.css" rel="stylesheet">
+
+<!-- versioned resources -->
+        <?php versioned_resource ('css/bootstrap.min.css'); ?>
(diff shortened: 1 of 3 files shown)


moodeaudio:
We versioned the files manually for 5 which is a real pain. 
	<link href="css/bootstrap.min.css?r50a" rel="stylesheet">

This PR automates it :-) There is already some code in playerlib that parses out the release number from footer.php

Btw, I added your userid to Test Team

moodeaudio:
Hi Paul,

I've integrated this PR into the upcoming r5.0c codebase.

-Tim

PR #31, opened 2019-05-20
Add missing parameter for ALSA

entroPyth:
Specify 0dB gain in addition to volume to help prevent clipping with default 100% state by adding " 0," in front of the 100 parameter specifying volume percent.

### DIFF
--- a/www/command/worker.php
+++ b/www/command/worker.php
@@ -563,7 +563,7 @@ workerLog('worker: - Miscellaneous');
 // since we initially set alsa volume to 0 at the beginning of startup it must be reset
 if ($_SESSION['alsavolume'] != 'none') {
 	if ($_SESSION['mpdmixer'] == 'software' || $_SESSION['mpdmixer'] == 'disabled' || $_SESSION['audioin'] != 'Local') {
-		$result = sysCmd('/var/www/command/util.sh set-alsavol ' . '"' . $_SESSION['amixname']  . '"' . ' 100');
+		$result = sysCmd('/var/www/command/util.sh set-alsavol ' . '"' . $_SESSION['amixname']  . '"' . ' 0,100');
 	}
 }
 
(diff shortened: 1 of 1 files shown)


moodeaudio:
Hi,

Here are some test results that indicate that the parameter represents left,right channel level% rather then a gain cap.

pi@rp2:~ $ amixer sset Digital 0,100%
Simple mixer control 'Digital',0
  Capabilities: pvolume pswitch
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 207
  Mono:
  Front Left: Playback 0 [0%] [-99999.99dB] [on]
  Front Right: Playback 207 [100%] [0.00dB] [on]
 
pi@rp2:~ $ amixer sset Digital 50%,100%
Simple mixer control 'Digital',0
  Capabilities: pvolume pswitch
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 207
  Mono:
  Front Left: Playback 104 [50%] [-51.50dB] [on]
  Front Right: Playback 207 [100%] [0.00dB] [on]

pi@rp2:~ $ amixer sset Digital 100%
Simple mixer control 'Digital',0
  Capabilities: pvolume pswitch
  Playback channels: Front Left - Front Right
  Limits: Playback 0 - 207
  Mono:
  Front Left: Playback 207 [100%] [0.00dB] [on]
  Front Right: Playback 207 [100%] [0.00dB] [on]

entroPyth:
Hey,
I don't have anything to test for digital signal, but it does seem consistent for analog.

moodeaudio:
I see from your Forum post that you are referring only to the Pi on-board audio device. This device is digital and its mixer name is "PCM". The example I posted is for a Hifiberry DAC+ which has a mixer named "Digital".

Obviously the 0,100% parameter does something different on the on-board audio device vs a DAC board like the Hifiberry.

Can you provide a link that explains the ALSA x,y volume param?

entroPyth:
Just restarted everything to double check findings and my suggestion isn't valid in reference to documentation. It looks like the incorrect syntax just happened to work in this single case.  It would appear amixer is offset by 4.00dB in reference to 100% at least for the on-board sound card of the raspberry pi 2 since specify by dB gets the intended result.  The rest is beyond my understanding, but it's been fun exercising curiosity to figure out more.

I believe you're right that this is a kernel issue responsible for +4dB at 100% as opposed an issue with the project code.  As a side note, I imagine there are a lot of people out there with RPis that think the on-board audio is sub par quality because of this.

moodeaudio:
Please submit the issue to the Pi Linux Git :-)
https://github.com/raspberrypi/linux/issues

If it's in fact a bug the kernel devs will fix it. If it's not a bug but "by design" then they will explain why its +4dB and not 0dB at 100%.

-Tim

PR #33, opened 2019-05-29
Fix position of random album button

chris-rudmin:


chris-rudmin:
There is a bug. I will reopen with a better fix

PR #43, opened 2019-07-22
Enable album art for iso files

epicfrequency:
Search upper directory for iso file cover art


moodeaudio:
How do I test this PR?

epicfrequency:
try an iso audio image. see if album art shows?

epicfrequency:
essentially if there is an iso file, it searchers in the upper directory instead.

moodeaudio:
Provide test data.

PR #44, opened 2019-07-26
Correct the placement of the Cue Icon

pccasto:
Not sure what the sx tag for images relates to.  Think this might make more sense if all of the images for db browsing were tagged with db-browse, but that is currently inconsistent.  If this (or something similar) is appropriate, then a lot of the 'float:left' in the javascript can be removed.


monkeypox:
Inline css should be migrated to the actual css files whenever it makes sense. There’s a huge glut of legacy css from bootstrap (mainly), so adding a few rules isn’t ever going to be a problem.

I don’t use the browse view so am not super familiar with it’s internals (like what the “sx” class represents) but it definitely could use some love if anyone ever felt like taking it on.

moodeaudio:
Hi,

I fixed this in a different way by adding the class .db-browse-icon {float:left;} to panels.css and then replacing the 4 style="float:left" statements in playerlib.js with the class.

Slowly working my way through the CUE handling PR...

-Tim

PR #45, opened 2019-07-27
Chrome on Windows 10 display issues in 5.3.1 (may have started with 5.2)

pccasto:
Chrome throws console error, and display is mostly blank when there is a stylesheet without href.

There is a bit of playerlib.js that iterates over stylesheets. At that point Chrome throws a console error, and most of the display is blank/unusable. Not sure why there is a stylesheet without href, but this code provides a defensive measure against that case.

Firefox did not have an issue.

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -3738,7 +3738,7 @@ function getCSSRule(ruleName) {
 	for (var i = 0; i < document.styleSheets.length; i++) {
 		var styleSheet = document.styleSheets[i];
 
-		if (styleSheet.href.indexOf('flat-ui.css') != -1 ) {
+		if (styleSheet.href && styleSheet.href.indexOf('flat-ui.css') != -1 ) {
 			var cssRules = styleSheet.cssRules;
 			for (var j = 0; j < cssRules.length; j++) {
 				var cssRule = cssRules[j];


moodeaudio:
How do I repro?

pccasto:
I did a clean install of 5.3.1 to get my system going again after the move.  Firefox on Windows 10 did not have any issues.Then I saw this problem while using Chrome running on a couple of different Windows 10 systems.

I just checked Chrome on Android - it does not have an issue with the original code. Perhaps Chrome on Mac will also silently ignore the issue? Or it iterates over the stylesheets in a different order, so that the one that does not have an href comes after the flat-ui.css, so is never seen?

The symptoms I was seeing was that the page would only paint the 'm' in the upper right corner, and the player control along the bottom.  The 'm' would allow for the configuration control popups, but nothing else, and the player control would not do anything. When using developer tools view, there was a console message:
playerlib.js?v=r531:3729 Uncaught TypeError: Cannot read property 'indexOf' of null
    at getCSSRule (playerlib.js?v=r531:3729)

When I revert my local patch and clear the cache, the problem reproduces with Chrome on Wndows 10.

moodeaudio:
Very odd. I traced the code and I'm not seeing how there could be a stylesheet defined w/o an href. Plus the function getCSSRule() definitely has to find the flat-ui.css stylesheet to be able to set the color of the radio button sliders used in the config screens, otherwise they will always be red.

1. Header.php defines calls to versioned_resource() for example versioned_resource('css/flat-ui.css');
2. Function versioned_resource() echos a line containing the stylesheet definition.
```
echo '<link href="' . $file . '?v=' . $_SESSION['moode_release'] . '" rel="' . $type .'">' . "\n";
```
I'd suggest reverting the patch and placing some logging as below.

```
// get css rule object
function getCSSRule(ruleName) {
	for (var i = 0; i < document.styleSheets.length; i++) {
		var styleSheet = document.styleSheets[i];
console.log(styleSheet.href);
		if (styleSheet.href.indexOf('flat-ui.css') != -1 ) {
			var cssRules = styleSheet.cssRules;
			for (var j = 0; j < cssRules.length; j++) {
				var cssRule = cssRules[j];
				if (cssRule.selectorText == ruleName) {
					return cssRule;
				}
			}
		}
	}
}
```

```
function versioned_resource($file, $type='stylesheet') {
	workerLog('<link href="' . $file . '?v=' . $_SESSION['moode_release'] . '" rel="' . $type .'">');
	echo '<link href="' . $file . '?v=' . $_SESSION['moode_release'] . '" rel="' . $type .'">' . "\n";
}
```

Then reboot and after a page refresh examine /var/log/moode.log and the Browser console. All the stylesheet defs should be logged in moode log and Browser console should show the first three: bootstrap.min, bootstrap-select and finally flat-ui.

Since I can't repro we should probably see if same issue occurs with moOde 6. I'll make an image today and PM you a download link.

-Tim

pccasto:
I had done something similar earlier with the console output, but hadn't gone into the playerlib.php.
I had to make some adaptations to get this to run, since the error (that I'm seeing) breaks the loop.

I'm still seeing a href=null in the console output, but nothing to explain it in the moode.log.
Got to run out for a while, but I'll do some follow-on troubleshooting this afternoon/evening.

~paul

moodeaudio:
I posted a thread in Test Team section of the forum that contains a link to r600 test image 1.

monkeypox:
This is possibly something along the lines of what necessitated specifying the css file we were looking for (though that ended up being a decent optimization too) - some browser extension loading a style sheet.

As far as I can tell Paul’s code just type checked the conditional so it should still function, it’d just bypass any rogue style sheets that were getting loaded into the environment.

moodeaudio:
You have a good memory. It's the old "rogue stylesheet loaded by Browser extension" issue.

@pccasto - I'll add the fix and then close the PR :-)

moodeaudio:
I used a slightly different test for consistency with other similar tests in the code.

```
typeof(styleSheet.href) !== 'undefined'
```

chris-rudmin:
@moodeaudio @monkeypox, Did you guys validate this fix? Just FYI that if the error is `cannot read property 'indexOf' of null`, then probably `typeof(styleSheet.href) === 'null'`
I personally prefer the truthy test and logical operators. EG `if (stylesheet.href && stylesheet.href.indexOf('flat-ui.css')) { ... }`

moodeaudio:
Good point. I'll run some tests.

moodeaudio:
Looks like I misinterpreted the error that was being reported.
```
Uncaught TypeError: Cannot read property 'indexOf' of null
at getCSSRule (playerlib.js?v=r531:3729)
```

You are correct, the truthy test that Paul proposed is what's needed here. It covers both "var is undefined" and "var = null" cases.

Also by setting stylesheet.href to null I was able to repro the console error and UI behavior being reported.

Nice catch :-)

PR #18, opened 2019-02-17
Handle response from listallinfo when data includes flac files with embedded cue sheets.

pccasto:
I initially noticed this as a problem with some of my discs not showing up when drilling down on the albums page.  Problem occurred in cases where multiple .flac files were in the same directory e.g. (disc 1).flac, (disc 2).flac.

Turns out that when responding to `listallinfo` mpd returns the data for the embedded cue/playlist as the last line of data for the file being processed. It does not add a final 'Last-Modified' line, as it would for a file-based playlist. The code that skipped a line after a directory or playlist line was skipping valid data.

This same problem (and approach for fixing) might apply to wavepack files with embedded playlists, but I don't have any of those, so limited the fix to just flac files.

I've included sample below that shows return from a directory with three flac files, and three external cue sheets.

Thanks again for MoOde!

```
directory: NAS/flac/Yes/The Ultimate Yes (2003)
Last-Modified: 2018-03-18T20:45:22Z
file: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 1).flac
Last-Modified: 2018-02-24T20:43:38Z
Time: 4769
duration: 4769.466
Artist: Yes
Disc: 1
Genre: Rock
Album: The Ultimate Yes
Date: 2003
playlist: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 1).flac
file: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 2).flac
Last-Modified: 2018-02-27T22:44:25Z
Time: 4767
duration: 4767.093
Album: The Ultimate Yes
Artist: Yes
Genre: Rock
Disc: 2
Date: 2003
playlist: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 2).flac
file: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 3).flac
Last-Modified: 2018-02-24T20:43:38Z
Time: 1211
duration: 1211.400
Disc: 3
Genre: Rock
Artist: Yes
Album: The Ultimate Yes
Date: 2003
playlist: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 3).flac
playlist: NAS/flac/Yes/The Ultimate Yes (2003)/Yes - The Ultimate Yes (Disc 1).cue
[...]

PR #46, opened 2019-07-28
Feature & bug fix related to Cue files and flac files with embedded cue

pccasto:
Closing out #18 and #19 in favor of this one. This rebaselines off of the 6.0 develop branch, so will be easier to test.

### DIFF
--- a/www/command/moode.php
+++ b/www/command/moode.php
@@ -329,6 +329,13 @@ else {
 					echo json_encode('OK');
 				}
 				break;
+			// LIBRARY PANEL
+			case 'listall':
+				// could allow for the no-path case, but only if needed..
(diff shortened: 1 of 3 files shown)


pccasto:
I'm still having a problem testing this, because I can't get the develop branch / 6.0 to run cleanly even without these changes. 

It may be because of the way I'm trying to run 6.0 (just changing out the var/www and var/local/www directories), but I think there is an issue with 'read_cfgs' command.

Will investigate more, and open an issue or PR if I spot any problems there.

moodeaudio:
Hi Paul,

I'm going to release moOde 6 and then get to this PR.

moodeaudio:
Will revisit the whole CUE handling thing when time permits or we get some additional dev help to look into it.

PR #54, opened 2019-10-15
Fix menu breakage on iOS 13 / A12 processors

monkeypox:
This works around a bug in webkit on iOS 13 that affects menus with blurred backgrounds where the blur coverage was missing for part of the menu.


monkeypox:
The update works around the bug in some other context menus by removing the white-space:nowrap rule, if anyone can figure out how to repro this on a simplified test case (e.g. http://jsfiddle.net/4yd1j3L9/) then let me know and I'll file a bug with Apple.

PR #64, opened 2019-12-03
Libary options for how Year labels are shown in both Tag and Album views

gbh-uk:
Tag view: information can be shown as a subtitle to the Tag list, as:
None, Artist, Year, Artist (Year).

Album view: Year information can be shown either before the album name, after the artist name, or not shown at all.

taglist_sub (subtitles in Tag list).  Default: 'Year'

show_year_av (show year in album view).  Default: 'Before Album'

+ fixed some minor HTML syntax in other, non-related, template files.


moodeaudio:
Woah, I didn't realize there were a bunch of minor HTML bugs :-0

I have no idea how those crept into the code base but I'll fix them for the 6.4.1 December bug fix  release and then get to the PR in January :-)

gbh-uk:
Hi Tim - looking through the changes as you're making them - did you overlook the </a> closing tag on row 30 of www/templates/net-config.html ?

moodeaudio:
Hi,

Here is what's in the dev branch for line 30

<a href="#net-config-defaults" data-toggle="modal"><button class="btn btn-medium btn-primary">Reset</button>

moodeaudio:
I see, missing the ``` </a>```

monkeypox:
Gbh could you post a screen sho

gbh-uk:
> Gbh could you post a screen shot showing the different options for tag view? The artist option in particular as I don’t really get it since there’s an artist column.
> 
> Well, I kinda do get why you’d want more info attributes there but maybe a different view style would be a better solution. I have some code I’ll submit later that changes the button bar interface to a menu which will make it easier to add different sections without worrying about screen real estate.



Hi, yes, there's an Artist column. But if the albums are sorted by year AND there's no Artist selected then the Album column becomes long list of album names and thumbnails. Leaving people to try to remember who's "Greatest Hits" it is, for example. Or who's the Artist for "Dreamscape" - for example.



The current 6.40 has added the year only.  So, again - I'm still trying to remember who released that particular album, but now I know it was released in 1999. Which is not a great help tbh.



Hence my proposal - why leave people to guess?  Just add the artist name before the year and then everything you want to know about an album is shown at a glance.  It's not taking up any more line-space.

Sure, if you click an artist in the Artist column then it becomes obvious. But with no artist selected it becomes a bit of a memory game....

Just my 2d.  But imagine trying to remember who the artists of the following albums are:

moodeaudio:
@gbh-uk,

Can you submit a new PR that just includes the Library enhancement and not all the other HTML fixes? I've already manually taken care of those :-) If there are more fixes to come then I think they should be separate PR's. This would make it easy for me to examine and merge.

I'm not that adept at Git so please understand if I'm off base.

gbh-uk:
@moodeaudio,

Not a problem. I should've kept them separate. I'm fairly new to Git and just bundled them all together for ease, which actually ended up complicating things a bit. Thanks for your understanding!
Any new PR's will be individual and specific from now on.

monkeypox:
@gbh,

Thank you. I agree the information can be useful, I’m just not sure that the album column of the tag browser is the ideal choice as it’s so space limited.

I’m admittedly a column view purist, and ultimately the decision is Tim’s, but I think the question to ask is what would be the ideal interface for how you want moode to present the data and maybe we can come up with a cool new library view style that displays things in a way that isn’t constrained by the tag browser format. If the tag browser as-is plus the enhancements to the album column is ideal for you though that’s ok too. ;)

gbh-uk:
@monkeypox 

Hi, actually I have no issue at all with the Tag browser as it stands, with additional information in the Album column -  precisely because it doesn't take up any additional space - two lines of text fit nicely next to the thumbnail.
I realise that some will prefer no info at all (purists like yourself), and others (like myself) the artist so I can see everything at a glance. That's why I put the 4 options together - none, artist, year, and artist (year).  Personally I'll only display the artist label but I wanted to give people the choice.
I posted some images on the forum and feedback seems positive to the proposal, but as you say, it's Tim's decision.
My purist tendencies are more towards hiding functionality that I'll never need or use - such as the 'random album' buttons, and the Google link, etc.  I'd love those to be optional too - perhaps I'll work on that next for me. But I appreciate that some people like random, shuffle etc.  Variety is the spice of life, so they say!

moodeaudio:
Is it Ok to close this since we have a new PR and the HTML bugs are fixed?

gbh-uk:
For sure, I'll close it now.

PR #65, opened 2019-12-05
'Year' labels in Tag and Album views - re-push

gbh-uk:
As per push #64 - without HTML corrections as requested.
Note: Need to work on an alternative for 'display:contents' in panels.css as this attribute isn't recognised in Microsoft IE or Edge.


moodeaudio:
Note to self:

taglist_sub (subtitles in Tag list). Default: 'Year'

show_year_av (show year in album view). Default: 'Before Album'

gbh-uk:
Re-submitted based on 6.4.2 baseline

PR #74, opened 2020-02-23
'Year' labels in Tag and Album views - re-basedlined for 6.4.2

gbh-uk:
Updated so as not to break the 'quick search' function for those experimenting with UI.tagViewCovers = false.

Note:

taglist_sub (subtitles in Tag list). Default: 'Artist (Year)'

show_year_av (show year in album view). Default: 'After Artist'

### DIFF
--- a/www/css/moode.css
+++ b/www/css/moode.css
@@ -291,7 +291,7 @@ li.modal-dropdown-text:focus {color:#eee;}
 	/* tag view */
 	.lib-entry {margin-right:0;}
 	.lib-entry-song {font-size:unset;}
-	.songyear {display:none;}
+	/*.songyear {display:none;}*/
 	/* folder view */
 	#folder-panel {padding-top:1.875em;height:calc(100vh - 1.875em);}
 	#database .btn {font-size:.9em;}
--- a/www/css/panels.css
+++ b/www/css/panels.css
@@ -183,8 +183,8 @@ img.coverart {border-style:none;border-radius:.1em;box-shadow:0 .1em .2em rgba(3
 .pll2 {font-size:.9em;margin-bottom:.75em;color:var(--textvariant) !important}
 #albumsList li, #artistsList li, #genresList li {line-height:calc(1.45em + 1vmin);}
 #albumsList .lib-entry img {width:3rem;margin:.15em .25em .15em 0;}
-#albumsList .album-name {display:inline-flex;vertical-align:middle;max-width:calc(100% - 3rem - .25em);line-height:1.25em;}
-#albumsList .album-year {display:none;font-size:0.85em;color:var(--textvariant);}
+#albumsList .album-name {display:inline-block;vertical-align:middle;max-width:calc(100% - 3rem - .25em);line-height:1.25em;}
+#albumsList .album-year {display:inline-block;font-size:0.85em;color:var(--textvariant); vertical-align:top;}
 .database .active {font-size:inherit;font-weight:700;}
 .playlist .active .pll1 {color:var(--accentxts);font-weight:700;}
 .tab-content.visacc .playlist .active .pll1 {color:var(--adapttext);}
@@ -299,7 +299,8 @@ code, pre {background-color:inherit;color:inherit;border:none;font-size:.8em;}
 #lib-genre {top:calc(1.5em + 1px);height:calc(100% - 1.5em);border-right:1px solid var(--btnshade);-webkit-overflow-scrolling:touch;box-sizing:border-box;}
 #lib-artist {top:calc(1.5em + 1px);height:calc(100% - 1.5em);left:33.33%;margin-left:0px;-webkit-overflow-scrolling:touch;box-sizing:border-box;}
 #lib-album {top:calc(1.5em + 1px);height:calc(100% - 1.5em);left:66.66%;border-left:1px solid var(--btnshade);margin-left:0px;-webkit-overflow-scrolling:touch;box-sizing:border-box;}
-#lib-album span {display:none;vertical-align:middle;}
+/*#lib-album span {display:none;vertical-align:middle;}*/
+#albumsList .artist-name {display:none;}
 #lib-file {overflow-y:scroll;overflow-x:hidden;margin:1px 0 0 0;height:100%;width:100%;-webkit-overflow-scrolling:touch;}
 #lib-coverart-meta-area {width:20vw;line-height:normal;padding:0;float:left;position:fixed;}
 #lib-coverart-img {margin:.5em .25em 0 .5em;}
--- a/www/js/scripts-library.js
+++ b/www/js/scripts-library.js
@@ -424,6 +424,9 @@ var renderAlbums = function() {
 	var output = '';
 	var output2 = '';
 	var tmp = '';
+	var tmpSub = '';   // For subtitles in Tag View
+	var tmpYear1 = ''; // For display of Year in Album View before Album title
+	var tmpYear2 = ''; // For display of Year in Album View	after Artist
 	var defCover = "this.src='images/default-cover-v6.svg'";
 
 	for (var i = 0; i < filteredAlbums.length; i++) {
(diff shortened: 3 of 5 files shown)


moodeaudio:
From what I can see the enhancement is to control how Year is displayed on line 2 underneath the cover in Album and Tag views. The sort/group by year is handled by the Library option "Album grouping".

Re:

1. The enhancement covers all the ways users have reported how they want to see (or not see) line 2 and year. Normally I avoid having too many options but this one seems pretty easy to implement.

2. Users often want more and more information displayed. For example a recent request to put a sample rate badge in the upper right corner of every cover :-0 The current Album view allows as many lines as needed below the cover. A lot of albums require at least three lines to display Album and Artist and some require more. It looks OK to me this way. If we reduce to a single ellipsis limited line under the cover don't u think users will complain? Is this a mobile-only config that uses a single line?

I still don't quite understand this one:
"a new option to add chronological sorting to the album column if an artist is selected and that’s currently labeled as “Album/Year"

If I set Library option to group albums by year and click an artist in tag view I see the album list sorted chronologically by year.

monkeypox:
> 1. The enhancement covers all the ways users have reported how they want to see (or not see) line 2 and year. Normally I avoid having too many options but this one seems pretty easy to implement.

Adding (year) to line 2 seemed like the consensus forum favorite and there’s usually more space on the artist line. I haven’t tried the patch but it seems a benign enough that it could just be the default and then maybe give an option to remove it for people who don’t want it.

> Users often want more and more information displayed. For example a recent request to put a sample rate badge in the upper right corner of every cover :-0 The current Album view allows as many lines as needed below the cover. A lot of albums require at least three lines to display Album and Artist and some require more. It looks OK to me this way. If we reduce to a single ellipsis limited line under the cover don't u think users will complain? Is this a mobile-only config that uses a single line?

Apologies to gbh for the off-topic... Two lines, title and artist (and year with this patch) but each ellipsis limited so it's only ever the two lines instead of spilling over as needed. Not like we should do something just because everyone else does it, but every other player does seem to do it, and it does look neater as you don’t get weird orphans (one or two word lines). The downside would be if you had a lot of albums with titles that are lengthy yet similar at the beginning, maybe classical albums would be affected more?

> I still don't quite understand this one:
> "a new option to add chronological sorting to the album column if an artist is selected and that’s currently labeled as “Album/Year"
> If I set Library option to group albums by year and click an artist in tag view I see the album list sorted chronologically by year.
> 

Group by year overall isn’t very useful (to me) because I don’t want all albums to be sorted that way, but it is nice if you select an artist from the artist column to then see those albums listed chronologically instead of alphabetically. Anyway the only reason I mentioned it was his Artist (Year) item was similar to what I use for the hybrid Album/Year grouping.

moodeaudio:
Ok I think I get it. Your thought is to have an option to sort the albums column by year in tag view when artist is clicked regardless of the sort option setting in Library config.

Looking at Amazon HD I see what you are talking about wrt the 2 line, width limited text under the cover. On desktop Amazon HD scrolls the text when hovering over it. I think we could probably do something this too.

Classical and Jazz titles and box sets might be troublesome for example the 3 album box set below. Depending on font size each album might appear as "The Man From Ipanema..."

The Man From Ipanema
Antonio Carlos Jobim

The Man From Ipanema [Disc 1]
The Man From Ipanema [Disc 2] - Instrumentals
The Man From Ipanema [Disc 3]

monkeypox:
> Ok I think I get it. Your thought is to have an option to sort the albums column by year in tag view when artist is clicked regardless of the sort option setting in Library config.

To be more precise I split view settings so tag and cover have their own independent view style, tag has the additional album/year as described and it only applies to album sort.

3B555444-7BBE-4784-BF7F-8D2C3115FC31

> Looking at Amazon HD I see what you are talking about wrt the 2 line, width limited text under the cover. On desktop Amazon HD scrolls the text when hovering over it. I think we could probably do something this too.
> 
> Classical and Jazz titles and box might be troublesome for example the 3 album box set below. Depending on font size each album might appear as "The Man From Ipanema..."

Right, that’d be the problem. Theoretically the album cover would be useful as the text is often redundant but on a box set it’s probably the same. We could show it on hover for desktop but that doesn’t help touch devices. It’s kind of pick your poison.

gbh-uk:
Hi. To answer the first question - no, this doesn't sort - it just optimises the information that's displayed to the user and takes up less space in doing so - particularly in the Album View by not giving it the first line all to itself (and pushing other, potentially more useful information, out of the block).

As for what information to display I agree it's all down to personal preference - I would always (and do) use 'Artist (Year)' on both tag and album views, but it costs nothing to give options. Plus I wanted to give the option for people to leave it like it currently is if they choose to.

However, with regards to sorting, it seems that we're working on something similar with regards to adding an Album/Year sorting option.  But I do it in one option to cover both views.

I (too) have an uncommitted push that I'm readying, based on a conversation with @owsla over the weekend.
Both he and I wanted to be able to see the albums in both the Tag and Album views sorted first by the Artist and then chronologically.  The difference (if I'm reading it right) is that we'd prefer the sorting shown *regardless* of whether an Artist is selected or not.

The driver?  Well, it was interesting, for example, to know that the earliest album I had in my collection was by Miles Davis - but seeing it first in the list was a bit odd.  And if I wanted to find 'The Kick Inside' by Kate Bush I had to remember that it was released in 1978.  So why not start with 'ABBA' (for example) and all their albums chronologically, then 'ABC' and their albums, and so on...

This is what it looks like (with no Artist selected):

Year-Artist-Tag

and

Year-Artist-Album

and it requires only one option which it applies to both covers both Tag and Album views:

Year-Artist-Option

Howver I'm open to feedback.  I admire your balance of matching the myriad of user requests I see on the forum with practical implementations.  Rob spends most of his time in Album view and I spend most of my time in Tag view - and this works for both of us.

gbh-uk:
I just noticed I commented out row 294 in moode.css while I was testing and forgot to re-instate it. So please ignore that change if you test.
Thanks,
Gary

monkeypox:
> As for what information to display I agree it's all down to personal preference - I would always (and do) use 'Artist (Year)' on both tag and album views, but it costs nothing to give options. Plus I wanted to give the option for people to leave it like it currently is if they choose to.

I think too much choice can lead to cognitive overload. I’ve mentioned it to Tim a time or two (lol) but one of the best things we can do to make moode more accessible to new users is to limit complexity and/or hide advanced or seldom used options from the majority of users who won’t care about them. Ease of use = more people use moode = more money comes in to help Tim keep the project going. This change is a good example, we could let the user choose artist, year or both, the order they’re displayed, etc. but using both is so low impact it seems pointless not to just do that with the option people liked best (and maybe have an opt-out setting though I’d like to see if there are complaints first).

> However, with regards to sorting, it seems that we're working on something similar with regards to adding an Album/Year sorting option. But I do it in one option to cover both views.

Unrelated to this patch but I really think we should split view prefs, what works best for album covers view isn’t always what works best for tag view and iirc it was very little code to do it.

Since 6.5 is an image only release it’d be a good time to pad out the sys_config table for future expansion (not that we should go crazy adding new preferences, or at least not ones people explicitly set in a dialog, see nagging above).

> The difference (if I'm reading it right) is that we'd prefer the sorting shown regardless of whether an Artist is selected or not.

I have the album column sorted by album as it’s an album column. :p Anyway that seems like a good use of Artist/Year, don’t forget to add some info text explaining it to the circle-i info element when you submit the pr.

moodeaudio:
I'm not sure what the benefit would be to adding reserved rows to cfg_system. If a feature needs a row it gets added via in-place update or when we build an image -only release. Extra unused rows add some overhead because the table is sent to the client on page refresh and every time MPD idle timeout occurs i.e. whenever any part of MPD state changes.

As far as this PR goes, we have several things going on:

1. Add options for displaying the year in various locations on the Artist line
2. Standardize on Artist (Year)
3. Maintain an option for hiding the Artist line in Tag view albums column
4. Add a new Album group/sort by Artist/Year for both Tag and Album views
5. Add a new Album group/sort by Year for just Tag view and only when Artist is clicked

Up untill now Tag and Album views have been in sync wrt group/sort etc., so user sees the same order and behavior of the Albums. 

We could do 1 and 3 or just as reasonably 2 and 3.
We could also do 4 or 5, or 4 and 5 but 5 would prolly need to be expressed as something like below.

"Album sort when Artist clicked in Tag view"
Default - Follow Album group/sort setting in Lib options
by Year

-Tim

monkeypox:
> I'm not sure what the benefit would be to adding reserved rows to cfg_system. If a feature needs a row it gets added via in-place update or when we build an image -only release. Extra unused rows add some overhead because the table is sent to the client on page refresh and every time MPD idle timeout occurs i.e. whenever any part of MPD state changes.

I thought before you were hesitant to add new session vars because of in-place updates? Maybe I misunderstood. I agree we could programmatically add new records as needed in the updater but null records in the table shouldn’t introduce any real overhead as they’ll just be ignored. 

> Up untill now Tag and Album views have been in sync wrt group/sort etc., so user sees the same order and behavior of the Albums.

The only thing that changes is the album column sort which was always sorted alphabetically prior to the new library code iirc. The connectedness, e.g. select an artist and go to covers and it shows only albums by that artist still remains.

> We could do 1 and 3 or just as reasonably 2 and 3.
> We could also do 4 or 5, or 4 and 5 but 5 would prolly need to be expressed as something like below.

I vote 2, 4 & 5. ;) I would always show the second line when the cover is present in tag view, and not otherwise. When I tested with tag covers the single line looked a little lonely next to the much larger cover; the second line with artist, etc. made it look more “complete”. I’m not sure how it’d look with an album where the title spilled into multiple lines though, probably worse.

As for 4 & 5, since they sort two different domains (artist & album) neither is really a replacement for the other. I agree Gary’s #4 being Artist/Year makes Album/Year more problematic as a name as they’re not identical in function but I’m not sure how else to label it. Album>Year? :p It’s a sort style though so if added should go in the sort menu rather than as an additional setting option imo.

FWIW Gary’s example on the forum here:

http://moodeaudio.org/forum/showthread.php?tid=169&pid=14868#pid14868

Illustrates the orphan issue with multi-line titles. It’s not that big of an issue in the grand scheme of things but it’s still something it’d be better not to have.

gbh-uk:
Hi - I agree wholeheartedly with the less-is-more approach proposed by @monkeypox.  
The config options I put forward in this PR for 'before' and 'after' etc. were basically so that existing users would see no change when moving to 6.5.0, but they would have a few options to change it.  However consensus on the forum seems to be for 2: Artist (Year), which kind of makes the other options a bit redundant.  I would vote to standardise on that and I would just need to reduce the config options to 'None' - hide the line in Tag view.

Incidently I've optimised the routines a bit since this PR as I remembered that there are two other cases we need to consider:  a. if the files do not have the Year tag populated then the line appears as **Artist (0)**,  and b. if the Year tag is not a number then it appears as **Artist (NaN)**.  So I've added code to strip those occurences out.

Mixing topics a bit : with regards to sorting - I'm pretty much ready with another PR for the grouping option "Artist/Year" that renders the Albums in both the Tag and Album views grouped by Artist with their Albums underneath them listed in chronological order.  This was how I've been looking for it to be for some time now, and how @owsla wanted them to appear.  This covers both points 4 and 5. I think?  My preference would always be chronological - but, as was mentioned above it's a case 'pick your poison'.

My proposed text for the (i) information button is:

> This option determines how albums are grouped in both Tag and Album views.
> Options are:
> 
> - "by Artist" - all artists, and the albums for each artist, will be sorted alphabetically,
> - "by Artist/Year" - all artists will be grouped alphabetically and the albums for each artist will be sorted chronologically beneath them,
> - "by Album" - all albums in the library will be listed alphabetically, irrespective of the artist,
> - "by Year" - all albums in the library will be listed chronologically, irrespective of the artist.
> 
> NOTE:
> When grouping by Year if not all tracks in the album have the same Year tag then the latest year will be used.
> If no Year tag exists or if the Year tag is not a number then no year information will be shown for that album.

gbh-uk:
I vote 2, 4 & 5 also - with no option to hide the Artist line in Tag view.
I can probably do an additional PR tomorrow so you can see how it would look and work.  
Quick question - would a PR for point 2. and a separate PR for 4/5 be better, or just a single PR that covers both?
They're implemented in quite distinct areas of code so it wouldn't be too difficult to distinguish the relative changes for both if they were in a single PR tbh...

moodeaudio:
I think I may need to mock up some code on my end to see what makes sense wrt 4 & 5.

@monkeypox, for #5 I'm still confused as to whether the sort options are truly separate for Tag and Album views. 

For example:

- Album view album sort = by Artist
- Tag view album sort = by Year (when clicking on Artist)

1. If no artist is clicked in Tag view does the album column default to the Album view album sort which in my example above is "by Artist"?

2. If an Artist is clicked in Tag view and given that Tag view album sort = by Year, what is the ordering of the albums after switching to Album view? Do they appear sorted by Year or by Artist?

moodeaudio:
Below are some screen shots after coding for a standardized Artist (Year) line.

Screen Shot 2020-02-26 at 8 27 20 PM
Screen Shot 2020-02-26 at 8 27 30 PM

monkeypox:
> Quick question - would a PR for point 2. and a separate PR for 4/5 be better, or just a single PR that covers both?

It’s better to keep things separate to make it easier for Tim.

> If no artist is clicked in Tag view does the album column default to the Album view album sort which in my example above is "by Artist"?

Yes. With the current code tag sort would be Album, cover would be Artist.

> If an Artist is clicked in Tag view and given that Tag view album sort = by Year, what is the ordering of the albums after switching to Album view? Do they appear sorted by Year or by Artist?

Cover shows sort by year as long as you’re in that artist selected context (tapping the header returns to regular sort prefs). I think we’d need to do a conditional re-sort on view switch if we want to maintain sort purity in that specific case but I need to look at the code to see what all is happening when a switch occurs as it may be possible to streamline things.

monkeypox:
New screens look good. :)

moodeaudio:
They don't look quite so good on mobile though cos too much text. This is where your proposed ellipsis limited lines might be the ticket.

I'm beginning to see the problem with the global Albums by year sort. We could simply eliminate it and go with just global sorts by Albums or Artists and always internally sort Artists by year. We would lose the full albums list displayed sorted by year but I don't think it's really that useful from the perspective of finding albums to play and quick index is disabled in this sort :-0

monkeypox:
> They don't look quite so good on mobile though cos too much text. This is where your proposed ellipsis limited lines might be the ticket.

This is what I use to limit album & artist to single lines, I implemented a simplified structure compared to 6.4.2 in the hopes of eking out a bit more speed where the lib-entry class is assigned to the li and divs were removed, but you can change the selector to apply it to whatever. If you want to just apply it to mobile we can hang a "mobile" class onto the #content (aka .tab-content) div and modify the selector to have #content.mobile first.

#albumcovers .lib-entry span {display:block;max-height:1.5em;overflow:hidden;white-space:nowrap;text-overflow:ellipsis;}

> I'm beginning to see the problem with the global Albums by year sort. We could simply eliminate it and go with just global sorts by Albums or Artists and always internally sort Artists by year. We would lose the full albums list displayed sorted by year but I don't think it's really that useful from the perspective of finding albums to play and quick index is disabled in this sort :-0

It'd be interesting to have metrics that showed what preferences people used. I think there's a place for Artist sorted alphabetically as well as by year though it wouldn't really bother me if it only showed by year.

PR #77, opened 2020-02-28
Sort and show Albums chronologically under Artists sorted alphabetically

gbh-uk:
Adds 'Artist/Year' album-grouping option.
Sort and shows Albums chronologically under Artists sorted alphabetically in both the Album column of the Tag view, and in the Album view.
Permits correct operation of Alphabits quick-search bar in both views (selecting by Artist).

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -2750,13 +2750,15 @@ $('#index-artists li').on('click', function(e) {
 });
 $('#index-albums li').on('click', function(e) {
     // .artist-name or .album-name
-    var selector = '.' + SESSION.json['library_album_grouping'].toLowerCase() + '-name'
-    listLook('albumsList li ' + selector, 'albums', $(this).text());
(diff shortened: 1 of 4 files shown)


gbh-uk:
Many thanks!

One thing - I can't seem to find the section of Github where you're sharing comments outside of PRs.  I'm receiving emails about [Library Work Items] but can't see how to take part in the discussion.

So until I find that functionality may I make one observation here....  I (personally) don't see why it would be necessary or useful to have different sort options for what is essentially the same information: namely one sort algorithm for the list of albums in the Album column in the Tag view, and another option to use a different sort algorithm for the very same Album information in the Cover view.  Would people not expect to see the same information sorted in the same way despite which view they're in? Or perhaps I'm misunderstanding the intent?

Anyway - thanks once again for merging my two PR's and I'll aim to be of service wherever I can.

moodeaudio:
Those conversations are in the Issues tab.
https://github.com/moode-player/moode/issues

moodeaudio:
I should add that the email should contain a link to the conversation/issue.

PR #82, opened 2020-03-02
library-year-filter

monkeypox:
Adds:

1) year filter so library search can return year based results by appending !r to the search query, e.g. "2019!r" returns albums with 2019 in their year tag, or "1970-1979!r" returns albums from the 70s). This has to be exposed in the library search function in scripts-panels but there are other dependencies beyond the scope of this pr.

2) split sorting for tag and cover views with new session var "tag_album_sort", this requires additional code in 1) playerlib.js and 2) footer.php but those are currently in flux behind the scenes so I don't want to add something that will conflict.

3) add album/year, will need additional playerlib & footer.php code (to set the option) to expose.

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -96,7 +96,8 @@ var NETWORK = {
 
 // Eventually migrate remaining global vars here
 var GLOBAL = {
-	regExIgnoreArticles: ''
+	regExIgnoreArticles: '',
+	ArtistSelected: ''
 }
 
 // live timeline
--- a/www/js/scripts-library.js
+++ b/www/js/scripts-library.js
@@ -32,7 +32,7 @@ var LIB = {
     currentDate: null,
     totalTime: 0,
     totalSongs: 0,
-    filters: {artists: [], genres: [], albums: []}
+    filters: {artists: [], genres: [], albums: [], year: []}
 };
 
 var allGenres = [];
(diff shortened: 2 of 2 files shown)


moodeaudio:
If I'm not mistaken the year search already works cos we are now carrying the year in the artist line.

1980 or (1980) returns all albums with that text in the Artist line which effectively is a Year search. Whats not quite there yet is a range. 198 works but it also catches Album titles that have that string. (198 should work but it doesn't for some reason.

monkeypox:
I saw that but it contains false positives because it matches with anything in the list.

moodeaudio:
True. It's the poor man's year search.

moodeaudio:
For number 2 it might be simpler to consider the full-album-list-sort for Tag view as an exception to the default sorts for the both Tag and Album views?

For example:

Default album sort [                ] (applies to both Tag and Album views)
by Album
by Artist
by Artist/Year
by Year

Tagview album sort [                ] (applies to the full Album list in tag view)
by Album
Default album sort

The settings below would achieve the particular sort that you are interested in where Tag view album column is alphabetic and clicking on Artist displays albums chronologically.

Default album sort [by Artist/Year]
Tagview album sort [by Album]

gbh-uk:
Probably I'm not the person to comment on whether people would find it useful or not to search for an album by year (or range of years) because I never use the search function at all as it stands. So it's never entered my head that search by year is a feature that's missing.  But the fact that swizzle has thought of it means that there's obviously a demand.  
I'm curious - would the '!r' suffix have to be remembered and entered by the user?

Re. 2 the only sort option that I'd ever use in both views is Artist/albums-by-Year.  
It suits me because I know that if I want to listen to an artist's first album then it'll be the first album in the list... and so on.  But equally I'm sure that there are people who prefer their albums to be listed alphabetically, and so it's catered for. 

Why not just stick with...

Tag view albums:
  by Album
  by Artist
  by Artist/Year
  by Year

Album view albums:
  by Album
  by Artist
  by Artist/Year
  by Year

...so people can pick their poison?
Then their choices are fixed and the sorting doesn't change from alphabetical to chronological depending on whether an artist is clicked or not?  Could be confusing and frustrating?

monkeypox:
> For number 2 it might be simpler to consider the full-album-list-sort for Tag view as an exception to the default sorts for the both Tag and Album views?
> 
> For example:
> 
> Default album sort [ ] (applies to both Tag and Album views)
> by Album
> by Artist
> by Artist/Year
> by Year

It’s more logical to give each view its own sort than to add a separate setting just to tag view. Maybe someone wants to sort tag by Artist and cover by Album or some other combination. I assume default for both would be Artist and users can work out what they like best for themselves.

> Probably I'm not the person to comment on whether people would find it useful or not to search for an album by year (or range of years) because I never use the search function at all as it stands. So it's never entered my head that search by year is a feature that's missing. But the fact that swizzle has thought of it means that there's obviously a demand.

We want our search to be smarter because search is the dominant form of interacting with the internet and that includes streaming services which are the dominant form of music consumption now. Anyway search is by far the fastest way of finding an album or albums by a particular artist.

> I'm curious - would the '!r' suffix have to be remembered and entered by the user?

Yes. Alternately we could trap the enter (e.key) instead, and I did that initially, but I was thinking of a future when we might be querying mpd for view contents instead of keeping a copy of the db in memory and we might want to target the search instead of just returning all results. Companies like google spend a lot of money and processing power trying to decipher search queries to return good results but that’s a deep rabbit hole. For example if the user is searching for an album titled “bad-search” and the user hits return because that’s what they’re used to doing then we have to determine intent. It probably wouldn’t be an issue for just this but if we extend search further then it gets complicated.  Anyway I was thinking of duckduckgo which uses a similar !whatever method to extend their search results (e.g. terminator !imdb). 

> the only sort option that I'd ever use in both views is Artist/albums-by-Year.
> It suits me because I know that if I want to listen to an artist's first album then it'll be the first album in the list... and so on.

Right but we can’t cater to just what we personally prefer to use. Sorting albums by chronological order if an artist is selected was a user request (iirc) that sounded more logical to me than just sorting the album column by year.

Since the user has to explicitly enable it as a different sort option I don’t think it’ll be all that confusing when the album list is sorted chronologically when an artist is selected, and cover view uses its own sort order if you switch to that view while an artist is selected so no confusion there either.

moodeaudio:
Let's assume Tag and Album view have their own sorts. It's also proposed that in Tag view, regardless of the sort, when artist is clicked the Albums column always appears sorted by year. For me that would be sensible/logical but users may want or expect the list to be sorted the same way as what was specified for the full albums list.

I think there may be three options

1. Album view sorts
2. Tag view sorts (for full album column list)
3. Album column sort when artist is clicked 
- default to full album column sort
- always sort by year

monkeypox:
Both year and artist/year already would sort it by year, so artist and album are really the only options the proposed change would apply to. In the case of artist wouldn't someone wanting to sort the albums column by artist & year already prefer to have artist/year enabled? So it’s likely only album where the user might want to apply the change and imo it’d be better off as an option in the tag sort menu rather than having a new library option.

moodeaudio:
>To clarify, in the pr the only time the album column is different is if album/year is the chosen sort type and an artist is selected.

Gotcha.

moodeaudio:
Number 2 and 3 are complete along with migrating the Lib options from PHP to JS. I'm liking the separate sorts for Tag and Album views and the convenience of being able to access the Lib options from the Appearance dialog. It's a nice enhancement :-)

@monkeypox, after I make this commit are u planning to do another PR for the !r search feature?

monkeypox:
> after I make this commit are u planning to do another PR for the !r search feature?

Yes, but that was tied into other code and I wanted to try to break the changes into smaller pieces so it wasn't so overwhelming to review.

moodeaudio:
Ok cool.

PR #83, opened 2020-03-05
Cosmetic knob changes

monkeypox:
These are a few cosmetic changes to jquery-knob js.

1) makes the knob background always a complete circle which looks more symmetrical.

2) makes the ends rounded

3) changes relative values to true from false, this means they'll adjust (or better adjust maybe is a better way to say it) in size when the window is resized or we send a window resize event.

4) adds settings for pixel offset & bloom effects settings, pixel offset is so the canvas container is larger than the knob otherwise the bloom would be cut off when it reached the edges of the container.

A more comprehensive PR with other UI changes will include the +/- button style as shown

knob no bloom

knob bloom

### DIFF
--- a/www/js/jquery.knob.js
+++ b/www/js/jquery.knob.js
@@ -747,6 +747,8 @@
                 r = (this.cv == this.v);
             }
 
+			if (sat == eat) {return;} // don't draw if they're the same = nothing to draw = no blob
+
 			if (this.o.fgColor.substr(0,1) == '#') {
 				var temp = hexToRgb(this.o.fgColor);
 				this.o.fgColor = 'rgb(' + temp[0] + ',' + temp[1] + ',' + temp[2] + ')';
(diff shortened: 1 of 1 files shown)


moodeaudio:
The inside +/- buttons are very close to the knob touch ring which could be trouble. The current volume knob has proven to be trouble free.

I'm Leary of changing the knob cosmetics in a way that would increase the likelihood of errant taps causing unintended volume changes.

monkeypox:
We could shift them in a bit, but this ain't my first rodeo ;)

button margins

I know you had reservations but I've been using software volume with this setup fairly regularly for a number of weeks to try to weed out problems and the button position hasn't ever been a problem. And the controls look so much cleaner this way... 

controls

moodeaudio:
Like @chris-rudmin said "don't be afraid to merge".

I'll try it out this week.

monkeypox:
Here's the div with the height changed to 70%, that gives some more room between buttons and knob.

Screen Shot 2020-03-05 at 10 05 43 AM

PR #84, opened 2020-03-07
ui updates

monkeypox:
This is the first take at migrating various ui updates.

1) viewswitch moves to a menu instead of button bar which is more flexible and reclaimed center space is used for descriptive text (e.g. view sort type, recently added, search results).

2) playbar changes: mobile overlays text onto a linear faded cover, this will probably come to regular, get rid of lozenge in favor of clicking on the whole playbar (need to check wrt voiceover).

3) switch from display to visibility (mostly) to toggle tab-content, this is massively faster with a large play queue but woe be unto he or she who tries to play their entire library while in chrome developer mode.

4) optimized potentially large list layouts, simpler = less memory and faster.

5) added range to library search (e.g. 1970!r or 1970-1979!r. a single year almost works as-is but can have false positives. currently this uses !r or just enter for testing, the latter would need validation especially if we expand what is possible with search but would probably be easier for the user.

6) changed the mv menu to a regular m with a ghetto motion blur effect instead because it looked weird with two v menus.

7) appearance modal remembers section state and disclosure triangles now work again. this needs a customize_state session variable.

8) added a potentially buggy font size option to appearance > other. this needs a font_size session variable.

9) mobile toolbar is dead, long live regular toolbar applied to mobile.

10) possibly controversially, and potentially buggy, coverview is now (mostly) toggled by adding a class to the playback panel, this lets us use regular controls, and gets rid of the duplicated playlist, and associated html, etc.

11) reduced library screen (and in general but primarily it impacted the library screen) cpu usage by removing all countdown timers except one and implementing a sync function that copies the value to relevenat elements when needed.

12) gets rid of pointless modal windows, makes the volume popup cool again.

13) moves the volume buttons inside the knob for cleanliness which is next to godliness.

This will undoubtedly have some bugs at this point.


moodeaudio:
"This branch has conflicts that must be resolved"

:-0

monkeypox:
Not to cast aspersions on git or the hub but the conflicts were like:

1) a comma was added to the regex global var to add other vars afterwards, resolved by putting it at the end so the offending comma could be removed.
2) some commented out code was removed, the bloat was replaced.

Then I merged develop in or something and now it seems to be ok but I zipped the directory prior to that just in case something bad happened.

moodeaudio:
Coolness. Thankfully we have been discussing these enhancements for many months and so most of the PR is very familiar to me :-)

I'm going to have to figure out how to resurrect the "Cosmetic knob changes" PR that I merged the other day but then reverted. I think I just need to revert the revert. Whatever u do don't delete the branch "library year filter" that contains those changes even if Github says it's OK to delete it.

monkeypox:
It’s safe on my computer. ;)

It just added rounded ends, the full circle background and potential for bloom so it’s not technically needed for the functioning of the pr.

Gary had a good idea in email or a pr note or something about how message headers were kind of opaque as a reset mechanism and he didn’t know at first to use them for that. We could track if an item is selected in all three lists and add an undo icon (used for single play repeat or something now) in the header so the user at least better knows something is different.

PR #85, opened 2020-03-08
Recently added items   highlight filter icon

gbh-uk:
The current 'recently-added-file' filter icon (Funnel) gives the user no visual indication as to whether the filter is active or inactive.
Also, counter-intuitively, clicking the icon turns on the filter, but clicking it again does not turn it off. The current functionality mixes the selection-reset click for the headers with the click to turn off the filter.

This PR adds:
1. an 'Off' appearance option to hide the Filter icon altogether for those who don't use or need it and prefer a cleaner look.
2. Fills and brightens the Funnel icon when the filter is active - to provide a visual indication.
3. Removes the filter 'turn-off' click from the Genre/Artist and Album headers and makes the Funnel icon a filter turn-on/turn off toggle.

Filter


moodeaudio:
Ya this would be a great update but I'm seeing "This branch has conflicts that must be resolved" message for three files.

:-0

gbh-uk:
I updated from the source before I added the changes and created the PR, but I must've missed something.
I'll give it another go again tomorrow!

monkeypox:
The filter is just for newly added albums? I moved that function to the viewswitch menu in my pr, the menu header reflects the context of the view (and the header no longer even exists in cover view). Examples with the cover view...

E0708D67-F1FB-44B0-B26B-5EAD718809BB

45A105FD-AAF1-494D-A0C3-38F5E8CE0956

2449612F-F0FC-4729-91AF-70F08E0B5F9E

gbh-uk:
> the menu header reflects the context of the view

That's much better.  Is this planned for 6.5.0 - in which case I don't need to fix the above conflicts?

The lop-config conflict is easy - the lop-config.php file simply doesn't exist anymore!

moodeaudio:
Right, no need to fix conflicts. I forgot that in monkeypox pr the button is moved to a menu.

PR #89, opened 2020-03-16
b2 fixes

### LINKED ISSUE #1: What the SSH user and password is ?
What about the SSH term server's default username and password is? I use the user from Pi is incorrect

monkeypox:
#1 migrate missing bootstrap.min.css changes

#2 fix radio cover size to use var(--thumbcols)

#3 revert 2x width for playbar cover

#4 use rem for playbar button margin/padding.

#5 add configurable extra metadata, currently the string is a global, it'll need ui in the appearance panel (other options?) and either need a new session var or some fancy coding

#6 commented out (probably) unneeded screen saver things

#7 added countup & countdown support for the playbar progress in synctimers

#8 use radiocover/radiolist in setcv instead of just radio (oops)

#9 change id to match css for the mobile volume popup "progress" bar.


moodeaudio:
Nice set of bug fixes. CV breakage is fixed :-)

I'll play with the extratags stuff.

PR #90, opened 2020-03-17
b2 fixes mk2

monkeypox:
#1 first steps at making config screens not do line splitting on mobile.

#2  adjust spacing for mobile wrt modals to better fit on smaller devices

#3 adjust how we display thumbs a bit wrt padding/margins

#4 fix font size on config screens

#5 add noscroll class to body when a modal is shown so the background doesn't scroll, this needed an edit of bootstrap.min.css so it needs to be re-minified for shipping versions.

#6 added a V button to minimize the playback screen on the top left

#7 modified how we display the m menu a little, also moved it in a touch

#8 inline-block made the year drop off the same line as the artist if it wasn't ellipsis limited so there are now separate display settings for both styles

#9 misc changes

### DIFF
--- a/www/css/moode.css
+++ b/www/css/moode.css
@@ -40,9 +40,8 @@ a.btn.btn-medium.btn-primary {font-size:16px;}
 .legend-configs {border:none;margin-bottom:.35em;}
 .legend-info-help {line-height:16px;padding-bottom:10px;font-size:14px;font-weight:normal;}
 /* general */
-.modal-backdrop{position:fixed;top:0;right:0;bottom:0;left:0;z-index:10000;background-color:var(--modalbkdr);backdrop-filter:blur(10px);-webkit-backdrop-filter:blur(10px);}
-.modal-backdrop.fade{opacity:0;}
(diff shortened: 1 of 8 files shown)


moodeaudio:
Nice :-)  I'm in process of reverting to the CV with Playbar that I posted in Comms Hub to see how that shakes.  I'll merge this PR in the coming days.

moodeaudio:
Merged :-)

5. questions observations
- Looks like u un-minified the file?
- I'm not seeing "noscroll" in the file
- we can just leave it un-minified unless there is a reliable tool to re-minify it

6. I'm not sure about the Vee button...

monkeypox:
I had to add the class add/remove and it was easier to do unminified. It’s not hard to minify but it’s not built into my editor like unminify is so I left it as-is.

Looking at the css diff the class was stopscroll not noscroll.

The v (as vee) was just a quick & dirty add. See if it grows on you, if so we can change it to playback-switch or whatever. It doesn’t annoy me much when it’s not front & center and it follows what most others do so I guess fewer people will be confused.

moodeaudio:
For sure lets play around with it.

PR #91, opened 2020-03-19
b2-fixes-mk3

monkeypox:
#1 modify cover size on shorter big screen mobile devices (e.g. Galaxy 5).

#2 adjust play control buttons a bit (goes with above to provide more space between cover, controls and text).

#3 revert coverview playlist to normal size to prevent it from possibly overlapping the knob(s).

#4 prevent knob changing if volume is disabled.

#5 disable mute if knob track is touched and then act on touch.

#6 ditto volume buttons.

#7 remove some deprecated ssvolume references I noticed.

#8 fix menu header to display "Albums by [artist name]" when an artist is clicked.

#9 fix missing space in ellipsis limited tag view between artist name and album year.


moodeaudio:
Merged :-)

I'll test out later today.

PR #101, opened 2020-04-30
Merge pull request #1 from moode-player/develop

gbh-uk:
User 'vinnn' reported an issue with spacing between Albums in tagview when appearance option Show tagview covers = No, here:

http://moodeaudio.org/forum/showthread.php?tid=2417&pid=19157#pid19157

Examination of the code also shows that the Artist (Year) information is still rendered, but then hidden with CSS.  Which is inefficient.

This PR proposal is to add an additional 'No [Show artist]' option - which un-hides these details (and spaces them).

And change the existing 'No' option to 'No [Hide artist] - (and spaces them) but in this case also removes the rendered HTML, thereby making the page more efficient.

ShowTagViewCoverOptions


gbh-uk:
Closed because it's now fixed.

PR #102, opened 2020-05-01
M Menu Hamburger

gbh-uk:
Simple PR to add a '3 bar hamburger' following the M menu; as a nod towards a 'traditional' UI and UX, and the users asking for it in the forum...

m-hamburger

### DIFF
--- a/var/local/www/header.php
+++ b/var/local/www/header.php
@@ -128,7 +128,7 @@
 
 		<!-- MAIN MENU -->
 		<div class="dropdown">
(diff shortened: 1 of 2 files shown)


gbh-uk:
If this gets implemented then the busy spinner should also be pushed a little more to the left: 
.busy-spinner rule in panels.css

Say right:4rem or a touch more.

moodeaudio:
A few users have been vocal about cues but most are super happy that they don't have to click the tiny lozenge anymore and can now have big click zones like the entire cover and Playbar :-)

I looked at a bunch of modern music apps and all of them had a learning curve that involved tapping around the UI to see what all the non-obvious elements did. 

Simplest might be to just display a first-use screen that was essentially Quick help but I'm not even sure that's really necessary.

PR #105, opened 2020-05-04
upgrade to upmpdcli-1.4.7

TheOldPresbyope:
### Prerequisites
As noted in the header to the COMPONENT 6 section, the following tarballs must have been downloaded to ~/moode/other/upmpdcli
- https://www.lesbonscomptes.com/upmpdcli/downloads/libnpupnp-2.2.1.tar.gz
- https://www.lesbonscomptes.com/upmpdcli/downloads/libupnpp-0.18.0.tar.gz
- https://www.lesbonscomptes.com/upmpdcli/downloads/upmpdcli-1.4.7.tar.gz

- all the existing tarballs and also the session.py patch should be removed

### Notes
1. the upmpdcli.service file referenced in the existing build recipe (`./moode/lib/systemd/system/upmpdcli.conf`) differs slightly from the counterpart in the upmpdcli-1.4.7.tar.gz; It seems to work but the differences should be noted.
2. The new upmpdcli.conf file distributed with the tarball may offer new functionality but the upmpdcli.conf file in moOde 6.5.1 appears to give all the functionality we currently use

3.The mediaserver function gives access to Tidal, Spotify, and Google Music. 
- Only the Tidal plugin has been tested.
- The Spotify plugin needs also the installation of libspotify. JFD provides a [link](https://www.lesbonscomptes.com/upmpdcli/upmpdcli-manual.html#UPMPDCLI-BUILDING)
- JFD says "upmpdcli Qobuz access does not work any more, so this is just kept as a historical note…​"
4. The OpenHome and Songcast capabilities have not been explored.

5. This build was tested on a fresh moOde 6.5.1 player. it is conceivable that it requires some dev library that should be explicitly identified but which was already present.

### DIFF
--- a/other/build/build_recipe_v2.21.txt
+++ b/other/build/build_recipe_v2.21.txt
@@ -11,6 +11,7 @@
 # - Compile miniDLNA 1.2.1+0763719f
 # - Image prep and build is Linux and Ethernet only
 # - Remove dead code and instructions
+# - Compile upmpdcli 1.4.7
 #
 # These instructions are written for Linux Enthusiasts
 # and System Integrators and provide a recipe for making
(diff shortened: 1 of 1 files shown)


moodeaudio:
Cool. Let's plan to include this in moOde release 6.6.0

moodeaudio:
I think we are ok to close this since 1.4.12 is going to be included in r660.

PR #107, opened 2020-05-06
Fix playlist alignment on small phone portrait

duracell80:
On a phone in portrait mode the calculation of the margin left on the second line of a playlist item is out of alignment due to an !important override. This fixes the issue for small phone screens in portrait mode.

Can be reproduced by adding one radio station to the playlist (clear/play). The line contains a microphone icon which the whole line is positioned slightly too far to the left (the calc isn't seeing the 3.8rem and is miscalculating). Adding the important override into moode.css fixes the issue. Will need to be included in mode.min.css to see the fix in the mobile browser.

Thank you for the great project to contribute to.


moodeaudio:
That appears to have fixed it :-)

PR #108, opened 2020-05-06
RPI 7" Fix padding on active playlist item and increase padding in playbar

duracell80:
On a small landscape screen such as the RPi 7" Touchscreen and a phone the active playlist item is intended slightly to the right. Adding padding of zero brings the active item back in line with the rest of the items for these screens.

The playbar at the bottom is also not sized well for the font size and elements are too close to the edges of the bar. Updating the RPi 7" screen resolves for that screen.


moodeaudio:
I'm not seeing the Playbar issues on my 7" Touch. The Playbar padding around the text looks ok and the previous button seems easy enough to see.

pi-touch

I've found that sometimes when viewing the Pi Touch 479 x 799 or 598 x 998 sizes in a viewport sim some of the elements are not positioned precisely like they are on a real Touch.

I am seeing the alignment issue in line 2 in the Playlist though but no the issue with active pl item.

monkeypox:
If this is relating to the ellipsis's not aligning for a song file vs streaming (aka without time) this is due to the width rule being commented out from the action element in a previous commit. The width needs to be set in order for css to know how to center the different styles.

There are a few things wrong with the playlist wrt alignment. 1) There’s some padding-left added to the active li, 2) the margin-left for .playlist span was changed from calc(3rem + 1vmin) to calc(3em + 1vmin) at some point and as the font sizes are different that leads to different positions on screen. This can be resolved either by changing the margin setting back to rem or setting the pll2 span to use font-size:1em. The advantage of the latter is it’d work better with the appearance level font size setting which would otherwise need to tweak the margin for large font sizes.

moodeaudio:
Ya for sure. I think the CSS is in an odd state...

duracell80:
The font size is larger which actually is really nice, but here's a compare between 6.4 and 6.5. I re-flashed 6.5.2. The bar looks a lot more cluttered and the play controls much less easier to see. Bringing back the light grey centered "tray lip" would help too. But yeah a lot less padding around everything in 6.5 it makes it a little harder to interact.

IMG_20200506_224510_copy_1008x756

monkeypox:
I don’t remember changing the padding on individual elements (though I’m not 100% sure what ended up in the release) but the older version of the bar is taller which would naturally look more spacious.

Looking at the older playbar css its height is 6.25em vs 4.8rem, something in the middle like 5.5rem looks good. The css for menu-bottom and playbar-cover would also need to be adjusted, and the playbar-title should have a larger number for line-height (like 1.75 vs 1.5). The taller bar would need additional adjustments wrt the indexes on the Pi though.

The contents in the old image has 5 columns which adds to the cleaner look, but that’s best addressed by letting users set the thumb size. The plumbing for that is basically done, it just needs some ui. It’d help to add a bit more space between the top toolbar thing and the div contents, it’s .25 now, .75 is probably a better choice.

The cover situation... the overlay is needed for mobile but not really landscape. If we keep it as-is changing the gradient start from (0,0,0,1) to (0,0,0,.65) or something to fade it a bit more into the background looks better to me. If we shift the play controls over to avoid the overlay there’s no avoiding them looking chunkier than 6.4 since they’re bigger than the buttons. Shrinking the font-size a bit, giving them more padding, and making the cover round (e.g. #playbar-cover img {border-radius:50%}) helps the chonk a bit.

I’ll post an image.

monkeypox:
B41023DC-AB41-43D1-B916-2F0E3FF98D82

moodeaudio:
I think the current Playbar size and layout is ok, particularly on the ghetto Pi Touch which is height challenged due to the regrettable choice of the designers to go with a 16:10 AR instead of the infinitely better 4:3 like an iPad Mini.

In any case I have a commit that includes some changes to moode.css that I think fixes most of the bugs. One anomaly I'm still not able to fix is a tiny amount of flattening on the bottom and right side of the knobs. This is only on the Touch. Btw after loading new CSS its often necessary to CLEAR the LocalUI browser cache via the option on System Config.

This may be complicated by the fact that my Pi Touch reports two viewport sizes depending on whether the Pixel aspect ratio was set to Default or Square.In the past it used to report only 799x479 :-0 Its discussed in this issue https://github.com/moode-player/moode/issues/109

Default: 998x598
Square: 998x553

duracell80:
I like the circle cover better than gradient there for sure.

Tim the knob "glow" shall we say I looked at the JavaScript in the canvas and the pixel offset isn't being taken into account on the height and width calculations. So if there is a pixel offset for that bloom adding the extra 12 pixels to the height and with and moving the x/y by 12 also brings the circle further away from the edges of the canvas.

Extra height in the playbar helps a lot, on my touchscreen the title of the current item appears outside of the playbar.

moodeaudio:
Which file to make the modifications in, and can you post or email me the changes? Thanks.

Heres a screen shot of Amazon HD desktop app which uses a circular thumbnail image. The play controls become visible after hovering over any part of the Playbar. 

Amazon HD

On your Pi Touch try clearing the Local UI browser cache. Its an option on System Config.

monkeypox:
We could also do something like this where we put a much smaller cover next to the track text and that lets the play controls hang out by themselves.

moode_(iPad Pro 11) (3)

This is the current bar:

moode_(iPad Pro 11) (4)

This is what setting the start of the gradient to .65 alpha looks like (which is what I use)

moode_(iPad Pro 11) (5)

moodeaudio:
What was that css for the darker gradient? I remember seeing it at one point a while back.

Option 1 could be neat but for short titles we would lose the balanced look of the centered title test.

We could also try setting the bg color of the previous button to same as the Playbar. It could make it stand out a bit better on light covers.

duracell80:
I'm gonna throw the wildcard in here in that most folks are right handed and even on a phone screen the bottom right is much easier to access than the bottom left.

Hover actions don't work well on touchscreens.

It was just the padding that was off for the meta data and scrubber line but the circle art does look cool. Take Amazon on at their UI game and punch out a dark circle in the middle of the art like an old style record so that the pause button stands out.

moodeaudio:
Amazon's mobile app has the play ctrls on the right like moOde mobile layouts. The lefty play controls are only on desktop/tablet, and hover of course is only on non-touch (desktop).

The other consideration that factors into whether or not to make changes is the amount of user feedback. If there were a lot of complaints about the Playbar controls and thumbnail then it would indicate that maybe some changes need to be made. This isn't cut and dry though because users that don't like something may not want to report it. We will never know.

I agree that we could improve certain layouts on the Pi Touch and make things bigger but mobile seems ok to me. I use an iPhone SE which is pretty small and Im still able to easily find and play stuff.

The API idea is interesting but who's going to maintain it and do end user support? Our project lives and dies by the quality of our end user support :-)

monkeypox:
I originally put the play controls on the right because of handedness but Tim liked the layout of the Amazon player so we flip-flopped them. I don’t use landscape moode all that often but when I have the controls on the left haven’t really bothered me.

moodeaudio:
Mobile portrait always has the play btns on the right :-)

duracell80:
User feedback is important for sure, there is some about elements being too small on mobile. One way to know about changes is a decline in adopting new versions or going back to older versions.

UX design can be invaluable though. I'm right handed on a touchscreen so the switch around of position now would make me less likely to use the playbar. Unless I moved my desk around to put the screen on the other side of the desk so the button are easier to reach.

I'm likely to stick with 6.4 for a bit.

moodeaudio:
We don't have the project bandwidth (or necessarily the desire) to track all that stuff but IME over the past 5 years of maintaining moOde there have always be segments of users on various versions including some that are still using the 4 series or even the 3 series. There are many reasons for this including but not limited to:

(1) the current version does not meet their needs because of feature changes or bugs
(2) they are happy with the version they have and don't want to expend the effort to upgrade
(3) they prefer an earlier version because they believe it produces better SQ
(4) they are experiencing upgrade fatigue

I'm also right handed and the lefty play controls on my iPad mini which has been on both the left and right side of my desk has never really been a problem.

PR #114, opened 2020-05-11
WIP:samplerate overlay

bitkeeper:
Example of the samplerate overlay. Not finished yet.


moodeaudio:
Do you think the hard coded media queries are generally applicable to most users?

If so then how can they be incorporated into the Appearance option "Thumbnail columns"?

monkeypox:
A few questions:

1) have you measured any performance impact of the patch, like before and after implementation? a small perf hit can add up when multiplied by thousands of instances.

2) wouldn’t an “hd” type badge accomplish much the same thing and also be potentially usable for tag view? I realize this scheme offers more differentiation but the number of users with multiple versions of the same music in differing high bit rate formats seems pretty small.

It’d also be helpful to attach an image of your work in action so we can more easily see what’s happening (even if you’ve posted on the forum as I haven’t been able to check there very often lately).

Additional thoughts:

1) if we use a class attached to the li of the relevant albums to denote hd-ness we could pretty easily add support so users could easily show just all their high-res music.

2) i wonder if there’s a way we could roll discovery of this info into the mpd update function or something and preserve the info so we only eat this particular bullet once... mpd stickers maybe? 🤷‍♂️

bitkeeper:
@moodeaudio no the media queries aren't part of the samplerate overlay, are just my personal tweaks. For clearity I will remove the queries from the commit.

www/css/moodecustom.css
```diff
@@ -0,0 +1,129 @@
...
+        --thumbcols:10.6vw;
+    };
+}
+
+/* pc screen fullhd+ ~12 cols */
+@media (min-width:1921px) and (max-resolution: 130dpi) {	
+	html, body{
+		--thumbcols:8vw;
+	}
+}
```
bitkeeper, line 70: This part has nothing to do with the samplerate overlay.

bitkeeper:
@monkeypox 
_1) have you measured any performance impact of the patch, like before and after implementation? a small perf hit can add up when multiplied by thousands of instances._

In my setup I don't feel any performance impact, however my lib is small, just below the 1000 albums. I expect data handling impact is very little, expect most inpact will be in the 2 added diffs with style applied.
If you have a standard procedure to measure the impact you would like me to run, just let me know ?
  
_2) wouldn’t an “hd” type badge accomplish much the same thing and also be potentially usable for tag view? I realize this scheme offers more differentiation but the number of users with multiple versions of the same music in differing high bit rate formats seems pretty small._

Personally I don't have multiple versions of the same album, I just keep the one that sounds the best to my ears. But what is HD ? For some CD red book quality is already HD. Only I just very like to see if something is 96k or 192k.

If one however don't like the overlays at all, it should be offcourse possible to turn of the overlays at all in the appearance options.

The implementation make a distinction in to 4 groups: lossy compression, 44k1/48k,  88.2k/96k and 176.4k or higher:
* Lossy compression - orange triangle is shown. no text. (I don't have this kind of files in my lib)
* 44k1/48k - red triangle + text shown
* 88.2k/96k - aqua triangle + text shown
* 176.4 or higher - yellow triangle + text shown

I'm not happy with the aqua and yellow colors, but it is good enough to demonstrate the intention of the overlay.

Most of lib albums is 44k1 and maybe consider to leave the overlay out voor 44k.1, but I can imagine that would be to specific for my personal situation.

At the top of PR is  image of the currrent implementation added and also a link to forum post with more proposals.

bitkeeper:
@monkeypox 

About additional thoughts:

_1) if we use a class attached to the li of the relevant albums to denote hd-ness we could pretty easily add support so users could easily show just all their high-res music._

I like that to. Tonight I will add a special class to the li.

_2) i wonder if there’s a way we could roll discovery of this info into the mpd update function or something and presoerve the info so we only eat this particular bullet once... mpd stickers maybe?_

I'm not sure I understand you?  The discovery is only done once at the backend side as part off cache creation function loadLibrary/genLibrary, just like the rest of the track information. It uses the LIBCACHE_JSON  as cache file. 

Or are you wondering about caching the albums info in addition to the trackdata in the backend, so the reduce function can be eliminated from the frontend ?

moodeaudio:
This is just a copy of my post http://moodeaudio.org/forum/showthread.php?tid=2202&pid=20070#pid20070 We should prolly just discuss here rather than two places. My bad.

MPD started to include the sample rate in its song file information but t's not a panacea. In my brief testing there were issues including DSF format showing no rate information and MP3 format showing a bit depth of 24 bit but of course there is no such thing as bit depth in lossy formats. I think in this case MPD is storing the decode bit depth.

Couple other points to consider when thinking about this as a generally applicable feature.

1. I think in addition to sample rate users would prolly want to also see the bit depth and format for example "24/44.1 FLAC" or "DSD 2.822 Mbps" same as what is displayed under the big cover in Playback panel. To do this requires executing the getEncodedAt($song, $outformat) function for each song file while the Library is being generated. This will prolly increase the time it takes to generate the Library. I won't know by how much until I test.

2. Displaying a badge with something like "24/44.1 FLAC" on each cover will prolly make the badge become the most dominant element in the panel instead of the cover art.  I don't think this is desirable.

monkeypox:
Marcel,
> If you have a standard procedure to measure the impact you would like me to run, just let me know ?
> 
I’m not sure about the php but you can approximate by doing something like:

https://stackoverflow.com/questions/313893/how-to-measure-time-taken-by-a-function-to-execute

By putting the first part before genLibrary is run then the console.log part after renderui somewhere though they’re both asymmetric execution so it’d be an approximation.
> But what is HD ? For some CD red book quality is already HD. Only I just very like to see if something is 96k or 192k.
> 
It’s definitely good info to have on a program level but I guess the question is what level of info is necessary to display and how to do it in a way that works for all library views. 
> Most of lib albums is 44k1 and maybe consider to leave the overlay out voor 44k.1, but I can imagine that would be to specific for my personal situation.
> 
What level of detail is necessary for you to be happy at the library view level? Ideally we want something that’s applicable to both cover and tag views.

> I'm not sure I understand you? The discovery is only done once at the backend side as part off cache creation function loadLibrary/genLibrary, just like the rest of the track information. It uses the LIBCACHE_JSON as cache file.

I didn’t look into the php but I assume it runs every time moode generates the library? The song files don’t change so this info could be theoretically cached when moode does the mpd library scan.

bitkeeper:
Probably found why MPD says MP3 is 24 bits; because MPD always decodes mp3 as 24 bit ( [source](https://wiki.archlinux.org/index.php/Music_Player_Daemon/Tips_and_tricks) ).

bitkeeper:
Summarizing till sofar:
* It seems that at least the common ground is that it would be nice to have that data (samplerate|bitdepth|filetype) available in the frontend.
* Opinions differ to what to do with it.
  * album thumb overlay with samplerate
  * album thumb text below with samplerate/bitdepth
  * show at album with tracks (in the library view)
  * filter on it; based on search or hd tag 
* Format information not correct for all audio formats:
  * mp3
    * Bitdepth of mp3 is not correct. Not a problem  I don't think anyone is using highres mp3 files ... ; -). 
    * With MP3 more likely you want bitrate as 320VBR instead of the samplerate.
  * dsd
     * Also with dsd your are interested in bitrate.
    * With dsd you don't care about classic bitdepth
  * Get the bitrate would require major changes because it needs mediainfo ( getEncodedAt($song, $outformat) )  to extract the data from each track.

bitkeeper:
Lets say that I would implement:
* a format field is added to loadLibrary
  * field contains three parts:  samplerate|bitdepth|file extension
  * mp3: set bitdepth to 16
  * dsd: leave the samplerate and bitdepth empty
* update the map reduce function for the albums support this field

Whould that make any chance to be accepted into the code based ?

At least you would then have a base to do something with the information. If someone would like to have bitrate for mp3 or dsd, is it to the next person to implement.

And second if I make a basic visualization with:
* Configuratable one of the 3 option:
  * no format info
  * text below thumb
  * album overlay
* set li class for hd (to prepare implementation of filtering)
  
Whould that make any chance to be accepted into the code based ?

moodeaudio:
Heres some data to chew on.

I tested generating the lib for my messy 664 track test collection (USB stick) that has a bunch of different audio formats and included the call to getEncodedAt() which shells out to mediainfo to extract the bit depth and rate for each track. Below is comparison of elapsed times with and without the call.
```
With --> 2 minutes (0.18 sec / track)
w/o:  --> 1 second (0.0015 sec / track)
```
It's a very expensive call. A 20K track collection would take about 1 hour to process. This would happen when first generating the library, regenerating the library or updating the library after adds or deletes. Unless there is a way to get a dramatic improvement in speed its prolly a non-starter.

It does produce super accurate data though. Heres a screenie showing the results. Impressive :-)

Screen Shot 2020-05-12 at 10 55 59 AM

So the question is whether to move forward with using MPD's format data which does not require shelling out and so should be fast but its not accurate in some cases or even present for others.

Options:
1. Use MPD format data, not worry about DSD, don't include bit depth for lossy
2. Offer both methods, fast and slow.
3. See if there are any PHP wrappers for mediainfo that don't shell out.

moodeaudio:
Another option for speed up might be to only extract the encoded_at data from the first track in each album.

bitkeeper:
mediainfo can also handle an entire directory (and recursive) or with wildcard. Formatting a reduced output also helps.

In this way instead of shelling out for each track, you can do it for each directory, album or what ever. Will not only save the overhead of shelling out, but also some startup of mediainfo itself.

1 file vs 12 files with mediainfo:
```
pi@moodedev:/mnt/NAS/losless/Deftones $ time mediainfo --Output=XML Deftones\ -\ White\ Pony\ \(2000\)/*.flac > /tmp/out.xml

real    0m0.548s
user    0m0.440s
sys     0m0.059s
pi@moodedev:/mnt/NAS/losless/Deftones $ time mediainfo --Output=XML Deftones\ -\ White\ Pony\ \(2000\)/01\ -\ Deftones\ -\ Feiticeira.flac > /tmp/out.xml

real    0m0.186s
user    0m0.151s
sys     0m0.030s
```

monkeypox:
Losing per track data would be sad for future moode but that enabled library build time is no bueno as-is.

What does your library screen look like with only the mpd data substituted in?

monkeypox:
Going back to what I mentioned before about having this run at mpd library scan time rather than moode library generation time, if we saved the info (mpd stickers, our own db, whatever) even if it was dreadfully slow if we only did it the one time we could have it run in the background on the pi and then invalidate the library cache when it’s finally finished so it gets regenerated and can fill in the now available bit depth / rate / etc. data.

moodeaudio:
Another option is to only use getEncodedAT() for DSF/DIFF and use the MPD Format tag for everything else. That should do the trick :-)

moodeaudio:
The strategy of using MPD Format tag as default and only using mediainfo for dsf/dff format works perfectly in genLibrary :-) I'll push a commit this morning.

- No bit depth is stored for lossy formats.I specifically test for 'f' in the bit depth element of Format tag or file extension 'mp3'. That seems to cover lossy. Example 44.1 MP3, 44.1 AAC.
- As a place holder for testing I'm just displaying the data underneath the cover.

As far as the feature goes Im thinking it would be something like below in the Library section of Appearance.

Display sample rate [                  ] ( I )
No
Text
Badge
HD only
( I ) help: The bit depth and sample rate used to encode tracks in the Album can be displayed in several formats on or below the cover art. The "HD only" selection displays an HD badge when the sample rate is greater than 44.1 kHz.

I'm still not sure about the "Badge" option because I think that having a badge on every cover will be very distracting but if we end up offering this then I'd propose using CSS to make the triangle instead of using images. I've tried some code but no success...

The badge for "HD only" option would prolly look similar to what Amazon does in its HD music app. User would click it to show the format info.

It's funny cos my collection is all 16/44.1 FLAC so this feature is not something I'll ever be using but I know that other users have some > 16/44.1 in their collection and so they prolly will like having it marked by an HD badge.

moodeaudio:
It took a bit of work but here is what I think will be a good basis going forward. I've coded it so that we can experiment with different badge styles, etc.

it looks like I do have some HD albums, 2 actually, lol

Screen Shot 1
Screen Shot 2
Screen Shot 3
Screen Shot 4

bitkeeper:
Great. And you even found 2 HD albums ? Then at least you don't do it for nothing ! LOL
Can I grab the code somewhere to give it a spin ?

moodeaudio:
I'm just finishing it up and will push the commit shortly.

bitkeeper:
Thanks, will pull it in the morning. 

btw for css triangles: https://css-tricks.com/the-shapes-of-css/ .  Works here fine including an opacity style with Firefox and Chrome.

moodeaudio:
You should be seeing the updated sources now.

New sql column is required.
```
sqlite3 /var/local/www/db/moode-sqlite3.db "INSERT INTO cfg_system VALUES (139, 'library_encoded_at', 'No')"
```

There is some code in scripts-library.js to prepare for displaying in Tag view but it's commented out. Not sure what to do in Tag view.

One unresolved issue is that the badge does not stay precisely positioned in the corner in different screen sizes. I'm not sure of the CSS to do this.

gbh-uk:
A cool addition.

wrt "The "HD only" selection displays an HD badge when the sample rate is greater than 44.1 kHz."

From Wikipedia "The term high-resolution audio is generally applied to audio with greater than 44.1 kHz sample rate **or** higher than 16-bit audio bit depth."

For example - I've many (so-called) HD releases that are 24-44.1 ...

moodeaudio:
Sure we could also test for bit depth.

moodeaudio:
I'm pretty happy with the last commit. It really streamlines the processing of encoded_at doing all the hard work on the server, and it provides an HD badge for the albums list in Tag view :-)

Lemmy know if there are bugs or anything else in this feature that needs attention.

PR #115, opened 2020-05-16
fix position of samplerate badge with long album names

bitkeeper:
**Issue:**
HD/text badge position not always correct:

image

**Precondition:**

- long album text (wider then album thumb)
- ellipsis limited text: NO
- Show sample rate: HD only or Text Badge

**Cause:**
- relative position from the botton, while the height can vary

**Fix:**
- make sure the batch div are just below the cover-menu
- changed style
- as bonus also added the text samplerate (#lib-encoded-at) to the big album view of the tagview


moodeaudio:
When I tested after merging the first time the position was off as shown in the attached screen shot so I reverted the merge. Then I realized that I had not installed all the files into my test system. After installing all the files though I was still getting the position issue so I reverted again.

I'm not quite sure what's going on but I am able to repro the issue where ellipsis text is NO. Would it be possible to do something similar to what @monkeypox did with the cover menu? It maintains its position regardless of the height of other elements.

Screen Shot 2020-05-16 at 10 02 34 AM

bitkeeper:
The construction is exactly like the covermenu That is why I need to move the dif. But think I know what gives the problem; didn't checkin the .min.js versions of the files. Did you generate them before testing ? (do you have a helper script for that .. or how do you do it by hand).

bitkeeper:
And the same for the .min.css files.

moodeaudio:
That was prolly the issue. I didn't re-gen the min's since they were included in the PR.

I use specific minifiers on my end so maybe it's best omit the .min's in the PR if possible.

Can u submit a new PR w/o min's?

bitkeeper:
The file .min include in this PR was www/templates/indextpl.min.html and that one is correct (edit by hand).
So no need to make a new PR I think ?

moodeaudio:
Yes, no need.

After doing 4th revert to get the PR changes back in and then regenerating the mins it's looking good now.

Nice fix :-)

gbh-uk:
Hi - just a suggestion but this feature adds 4 calculated and rendered divs to the page markup, regardless of whether a user enables the feature or not.

As it's something that a user is likely to not want, or will 'set once and forget' would it be better and lighter on the page if the divs were wrapped in a conditional and the library re-rendered if the option is changed, rather than always rendering all 4 divs and then hiding them?

That way the page will add either no divs or two divs to the markup.

moodeaudio:
Makes sense. I'll have a look.

moodeaudio:
I have the conditional version working but the question is whether to always have a div that contains encoded_at even if user sets the display option to No. Doing this would allow search filtering by sample rate for example 24/96 or by format for example MP3.

bitkeeper:
I would like the additional div, but then al mine web clients are powerfull enough.
Alternative 1 @monkeypox  suggested to add only hd class to li it self.
Alterantive 2 only searchable when a Show sample rate option is chosen.
Alternative 3 Add a new option to the drop down list of Show sample rate:
*No 
*No (but searchable)
*Text
*Badge
*HD Only

moodeaudio:
The "No (but searchable)" option is doable. I'll mock it up.

moodeaudio:
I pushed the commit that does the following:

- Converts from show/hide to dynamically added divs during lib load
- Changes from text in the sql col library_encoded_at to number for more efficient conditional in the renderAlbums() for loop.
- Adds option "No (searchable)" to the dropdown
- Defaults to adding a hidden div containing "bits/rate format" for the options "No (searchable) and "HD only".

bitkeeper:
Did pull and checked pc with chrome on (flac and some mp3):
* album view
 * all 5 settings 
 * include ellipsis limited text with (hd) badge position
* tag view
 * 2 modes hd badge and no
 * encoded_at album cover with tracks overview
Looks good! Phff

PR #116, opened 2020-05-17
Proposed Library Menu enhancements

gbh-uk:
Proposed enhancements to Library menu, as described in forum post: http://moodeaudio.org/forum/showthread.php?tid=2540&pid=20233#pid20233

### DIFF
--- a/www/css/moode.css
+++ b/www/css/moode.css
@@ -97,8 +97,8 @@ li.modal-dropdown-text:focus {color:#eee;}
 .help-block, .help-inline {/*color:var(--textvariant);*/color:inherit;}
 .help-block-modal {margin:0 !important;}
 /* dropdown menus */
-.dropdown-menu > li > a {line-height:2.75em;margin:0;padding:0 .5em;background:none;color:var(--adapttext);font-size:1rem;white-space:nowrap;}
-.dropdown-menu {float:right;position:absolute;min-width:200px;left:auto;border:none;list-style:none outside none;background-color:var(--adaptmbg);box-shadow:0px 4px 10px rgba(0, 0, 0, 0.4);border-radius:.5em;backdrop-filter:blur(10px);-webkit-backdrop-filter:blur(10px);z-index:1000;white-space:normal;}
+.dropdown-menu > li > a {line-height:2.75em;margin:0;padding:0 .5em;background:none;color:var(--adapttext);font-size:1.1rem;white-space:nowrap;}
+.dropdown-menu {float:right;position:absolute;min-width:260px;left:auto;border:none;list-style:none outside none;background-color:var(--adaptmbg);box-shadow:0px 4px 10px rgba(0, 0, 0, 0.4);border-radius:.5em;backdrop-filter:blur(10px);-webkit-backdrop-filter:blur(10px);z-index:1000;white-space:normal;}
 .dropdown-menu.open {padding:0;margin:0;}
 .bootstrap-select .dropdown-menu {background-color:transparent;}
 .btn-group.open.bootstrap-select .btn.dropdown-toggle {background-color:rgba(128,128,128,0.3);box-shadow:none;border-radius:4px;}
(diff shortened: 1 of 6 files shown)


moodeaudio:
Is this just the check icon menu or are there more changes?

moodeaudio:
After looking at the diffs I see that this PR includes multiple UI changes. I'd prefer that it be broken into 2 or more separate PR's. The first should be a PR that converts the existing Library menu to a "check icon menu" which is prolly an easy win. 

The rest should prolly reflect individual UI changes rather than bundle multiple changes into a single PR.

gbh-uk:
Hi Tim. It contains exactly what I described in the forum post:

i. Check icons in the menu
ii. Additional menu item 'Playback' and making the menu visible on the Playback panel
iii. And the title of the menu changes dynamically to reflect the current panel.

Do you want me to strip it back to just i. ?

monkeypox:
Adding a playback item to the library view menu is bad ui because the playback screen is not a library view. The playback screen is conceptually the expanded version of the play bar. If you go this direction you may as well add the configure screens, etc. and have one giant menu. 

I purposely didn’t add the view name in the library menu because 1) it’s evident by the view itself, 2) the menu header usually says what it is, 3) the menu header often carries useful information other than what view is showing.

The playback screen is uncluttered on purpose, a caret on the top-left of the playback screen (a la Tidal, etc.) was added by me then removed by Tim. If you do anything I think that is the least bad way to go.

moodeaudio:
I'd like to test check icons in the menu but the other changes sort of conflict with the existing design as @monkeypox has described.

lol, I remember that caret now but Im not sure why it got thrifted. Maybe I was more OCD than usual.

gbh-uk:
_Adding a playback item to the library view menu is bad ui because the playback screen is not a library view. The playback screen is conceptually the expanded version of the play bar._

Conceptual or not I feel it would be helpful to acknowledge that it's also the first 'view' a user is presented with when installation is finished.  They probably don't think of the screen as being a 'conceptual' extension of something that they haven't seen yet and probably wouldn't find it odd to see a menu there to help them switch out of it.
I'm puzzled though how, what is on the face of it, a reasonable and minimal change, intended to improve the overall UX and avoid users from having to come to the forum to report that they installed the ISO 5 times because they couldn't figure out how to leave the first screen and asking how to do it, can be described as 'bad-ui'.  
Perhaps adding a caret will suffice. My take on it was to unhide the Library menu so that the user interface became consistent across all screens.  Both are valid I guess.

_If you go this direction you may as well add the configure screens, etc. and have one giant menu._

That would be illogical.  One menu is related to switching between panels (conceptual or not). The other is an 'm' that leads to the config items.  No-one was suggesting to combine the two or that the UI or UX would be improved by doing so.

The menu-header is another topic.  Like I said I left it untouched.

A bit of background - I found moOde in September on my search for a Pi based music player, and almost ditched it because of something really trivial: the song time counted down as the song played, and not up.  It took me a while, exploring the various screens, to find out that it was indeed possible to change it to counting up.
 
That's a trivial example I know.  But counting up instead of down was a deal-breaker for me.  
So I can just imagine how many potential new users may not bother reinstalling 5 times and give up on the first screen...  I'm glad I wasn't hasty because I truly believe moOde is the best there is and I'm glad that I'm able to offer some small nuggets of assistance from time to time.

I get it however. I'm CTO of a company that designs and develops software applications for the medical industry. It's human nature that my development team is defensive (protective) of something they had a hand in designing, and I'm certainly not criticising the current UI.  I was merely suggesting with this PR that there's always room for minor enhancements that could help alleviate some of the frustration that users report.  It's probably a nervous tick I've developed in my job trying to keep doctors happy.  Especially in the times we're all currently living through.

I'd invite you to give the PR a whirl as it stands.  It's not a drastic change and isn't a million miles away from where the interface is right now.

But I'll also move on and strip the PR back to just the the check mark icons as Tim has asked.

Cheers,
Gary

moodeaudio:
We already have several proposals from users for addressing the first use scenario. 

1. Adding the lozenge or caret back in
2. Adding a first use screen that explains what the non-obvious elements do
3. Adding removable cue's to the non-obvious elements

I'm not opposed to doing any of these or if there is something better then doing that.

PR #127, opened 2020-05-23
In the function getCSSRule remove hard dep to 'flat-ui.min.css'.

bitkeeper:
`playerlib.js:getCSSRule` uses a hardcoded stylesheet name in the code, prob done for search optimizing. This construction is in the way of optimizing the resource loading or developing without minimizing.

moOde only calls  `getCSSRule` 3 times and only during startup. For now lets remove the optimization, by removing the `if`. No noticeable difference in loading speed.

```javascript 
// Get CSS rule object
function getCSSRule(ruleName) {
	for (var i = 0; i < document.styleSheets.length; i++) {
		var styleSheet = document.styleSheets[i];

		if (styleSheet.href && styleSheet.href.indexOf('flat-ui.min.css') != -1 ) {
			var cssRules = styleSheet.cssRules;
			for (var j = 0; j < cssRules.length; j++) {
				var cssRule = cssRules[j];
				if (cssRule.selectorText == ruleName) {
					return cssRule;
				}
			}
		}
	}
}
```

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -2848,13 +2848,11 @@ function getCSSRule(ruleName) {
 	for (var i = 0; i < document.styleSheets.length; i++) {
 		var styleSheet = document.styleSheets[i];
 
-		if (styleSheet.href && styleSheet.href.indexOf('flat-ui.min.css') != -1 ) {
-			var cssRules = styleSheet.cssRules;
-			for (var j = 0; j < cssRules.length; j++) {
-				var cssRule = cssRules[j];
-				if (cssRule.selectorText == ruleName) {
-					return cssRule;
-				}
+		var cssRules = styleSheet.cssRules;
+		for (var j = 0; j < cssRules.length; j++) {
+			var cssRule = cssRules[j];
+			if (cssRule.selectorText == ruleName) {
+				return cssRule;
 			}
 		}
 	}


moodeaudio:
I tested an instrumented version of the function and there's a pretty big diff between the number of iterations with the conditional vs  without the conditional.

with = 177
w/o = 1003

```javascript 
function getCSSRule(ruleName) {
    var k = 0;
	for (var i = 0; i < document.styleSheets.length; i++) {
		var styleSheet = document.styleSheets[i];
		if (styleSheet.href && styleSheet.href.indexOf('flat-ui.min.css') != -1 ) {
			var cssRules = styleSheet.cssRules;
			for (var j = 0; j < cssRules.length; j++) {
				var cssRule = cssRules[j];
                k++;
				if (cssRule.selectorText == ruleName) {
                    console.log('totals',k+i);
					return cssRule;
				}
			}
		}
	}
}
```

bitkeeper:
I don't think that the performance increase isn't such a big deal. See the results below ;-)
Espcialy because hardcoding against stylesheet name in the javascript introduces a maintenance nightmare in the end.

Using the original 'optimized function':
```javascript
console.time('getCSSRule');getCSSRule('.toggle .toggle-radio');console.timeEnd('getCSSRule');
getCSSRule: 0.255615234375ms
```

Using the 'unoptimized if removed function':
```javascript
console.time('getCSSRule');getCSSRule('.toggle .toggle-radio');console.timeEnd('getCSSRule');
getCSSRule: 1.1982421875ms
```

If you want an optimized version, just replace the getCSSRule by a thirdparty specialized library like styler (https://github.com/tjcafferkey/stylerjs/):
```javascript
console.time('someFunction');styler('.toggle .toggle-radio').set({'backgroundImage': 'url("'+img+'")'});console.timeEnd('someFunction');
VM22514:1 someFunction: 0.257080078125ms
styler: 0.2568359375ms
```
This includes changing the style attribute is self, the img was filled with the svg.

moodeaudio:
True about hard coding. 

IIRC there was another version of that function that we tried that didn't do any looping or have any hard coded names but it had some sort of issue with Firefox way back then. 

I'll have to do some digging.

monkeypox:
Iirc version...

#1 used an arrow function older tablet browsers didn’t like.
#2 removed the arrow function but would throw a cross site error when a browser extension included a css file (and this would likely happen with #1 as well)
#3 added the hard coded file name.

If #2 isn’t an issue 1ms wouldn’t be noticeable

moodeaudio:
Right, the CORS issue. It was fixed in the 5.2.0 release a year ago and then the function was subsequently optimized in later release.

Couple questions:
 1. The radio-toggle class which is the one being updated is only used in the Config screens so can we eliminate the call to getCSSRule in playerlib and scripts-panels?
2. Could we replace the function with a JQuery $().css() approach?

Here's the original function.

```javascript
function getCSSRule(ruleName) {
    ruleName = ruleName.toLowerCase();
    var result = null;
    var find = Array.prototype.find;

    find.call(document.styleSheets, styleSheet => {
        result = find.call(styleSheet.cssRules, cssRule => {
            return cssRule instanceof CSSStyleRule && cssRule.selectorText.toLowerCase() == ruleName;
        });
        return result != null;
    });
    return result;
}
```

Issue #129, opened 2020-05-23
Building moOde with Gulp


bitkeeper:
[Gulp](https://gulpjs.com/) is a toolkit to automate and enhance the workflow of webapps.
For C-minded see it like Make or Java troops just like ant.

Gulp requires npm to be installed.

At the moment I'm working on two different use-cases with gulp:
* Building the frontend webapp:
  * minify every js and css
  * no need to checkin intermediate files like .min.js|css anymore
  * bundle all the js and css file to one js and one css file (going 27 resources to 2 resources).
  * generate sourcemaps for the browser to still make easy debugging possible
  * automatic generate header or update header of files with the current date and or version info
  * deploy a build (like ~pi/moode/distr) to the final location (like /var/www)
  * automatic apply css browser prefixes, so you don't have to
  * make the use sass for css possible
  
* Make it possible to do frontend development work on my pc
  * use the development tools of your flavor on your PC
  * perform a build of the webapp
  * start a local test webserver (exludes php)
  * which include a proxy to your moode instance for the backend calls
  * optional automatic reload browser on code change
  * it will be more in the comfort zone of a webdev
 
If you are interested and or have feedback pleas join the discussion.

bitkeeper:
Before use:
* `install npm`
* open a command prompt
* cd to your work git dir of moode
* run `npm install` to get the dependencies.
* run `gulp` with `build|watch|deploy` or use your favorite editor/ide with gulp support (VS, webstorm etc)

The buildfile is gulpfile.js, the project parameters are in package.json.

Commands:
```
gulp build                - Build frontend from {app.src} and put output in {app.dest}

gulp watch [--production] - Start web server with as root {app.src} with proxy to moode, used for local web development. 
                                            When --production is given, also perform a build and use {app.dest} as web root

gulp deploy [--test]      - Deploys everything needed (inc php etc) {app.deploy}.
                                       With the option --test deploy to {app.dist}, default build/dist to check the 
output of an actual deploy.
```

Typicalcommand used on the pi:
gulp build
gulp deploy --test
gulp deploy

Typical use on the PC:
gulp watch
gulp watch --production
gulp deploy --test


If the buildin webserver is very slow, your hitting a know issue with local networks that have .local as domain. In that case just add to your hosts file moode.test with the specific ip number.

moodeaudio:
I'll get to looking at this after I finish the new Radio feature which will take me some number of weeks worth of effort.

bitkeeper:
[...]

PR #130, opened 2020-05-24
Support for frontend build/development with gulp. 

bitkeeper:
Example of gulp support for building and developing the moOde javascript/css frontend.

See issue #129  for more info.


bitkeeper:
Full agree with you. It makes sence to transpilefrom ES6. In recent user feedback some older device where already broken due ES6 code.

www/footer.php
```diff
@@ -185,8 +187,6 @@
 
     //workerLog('-- footer.php');
     $return = session_write_close();
-    //workerLog('session_write_close=' . (($return) ? 'TRUE' : 'FALSE'));
+	//workerLog('session_write_close=' . (($return) ? 'TRUE' : 'FALSE'));
+	echo "</body></html>";
```
bitkeeper, line 191: The opening of the html and body is done in the header.php. By moving the closing of the html and body from html to php, the source document can be parsed again by regular minifiers and validators .
Change nothing to the functionality.

www/templates/apl-config.html
```diff
@@ -16,7 +16,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  *
- * 2019-05-07 TC moOde 5.2
+ * @@__CURRENTDATE__ TC moOde @@__VERSION__
```
bitkeeper, line 19: Changed one source to demonstrated use central setting for auto replacing of current date and version in deployed files.

www/header.php
```diff
@@ -39,24 +41,47 @@
 
     <!-- RESOURCES -->
     <!-- Common CSS -->
+	<!-- build:css css/styles.min.css -->
```
bitkeeper, line 44: the build and enbuild tag are use to extract which resources are used and to which bundle it should be added.
A new header.php is generated which contains the bundles,

www/header.php
```diff
@@ -23,12 +23,14 @@
  *
  */
 -->
+<!--removeIf(GENINDEXDEV)-->
```
bitkeeper, line 26: Some comments are added which are used for preprocessing if generating a frontend only development index.html.
Doesn't break anything.

www/header.php
```diff
@@ -204,3 +242,13 @@
     <div id="cv-playlist">
         <ul class="cv-playlist"></ul>
     </div>
+
+	<!-- Only included when generate index.html for developmed purpose -->
```
bitkeeper, line 246: When generating a development only index.html it should also contain the content out of index.tpl.html and the footer.php.

www/header.php
```diff
@@ -204,3 +242,13 @@
...
+	<!--=include templates/indextpl.html -->
+	<!--=include footer.php -->
+
+<!-- make wellformed html; correct unclosed body and html (normally done by footer ) -->
+<!-- GEN_DEV_INDEX_TAG
```
bitkeeper, line 251: To make html precocessing etc possible like the bundle generation, the html should be wellformatted

PR #131, opened 2020-05-24
DEMO:If supported, use native lazy loading of the browser.

bitkeeper:
Demonstrates the use of native lazy loading in combination with fallback if the browser doesn't support it.

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -1264,7 +1264,7 @@ function formatBrowseData(data, path, i, panel) {
 			if (data[i].file.substr(0, 5) == 'RADIO') {
 				if (panel == 'radio_panel') {
 					var imgurl = '../images/radio-logos/thumbs/' + filename.replace(path + '/', '') + '.jpg';
-					output += '"><div class="db-icon db-song db-browse db-action"><img class="lazy-radioview" data-original="' + imgurl  + '"><div class="cover-menu" data-toggle="context" data-target="#context-menu-radio-item"></div></div><div class="db-entry db-song db-browse">';
+					output += '"><div class="db-icon db-song db-browse db-action"><img class="lazy-radioview" loading="lazy" data-original="' + imgurl  + '"><div class="cover-menu" data-toggle="context" data-target="#context-menu-radio-item"></div></div><div class="db-entry db-song db-browse">';
 				}
 				else {
 					output += '"><div class="db-icon db-song db-browse db-action"><a class="btn" href="#notarget" data-toggle="context" data-target="#context-menu-radio-item"><i class="fas fa-microphone sx db-browse db-browse-icon"></i></a></div><div class="db-entry db-song db-browse">';
@@ -3201,32 +3201,49 @@ function setLibMenuHeader () {
 	$('#menu-header').text(headerText);
 }
 
+/**
+ * Enable lazy loading of images for a certain view.
+ * Instead of src the image should have a data-original.
+ * 
+ * When the client doesn't supports native lazy loading the library jquery.lazyload is used.
+ * 
+ * @param {string} view - the view for lazy loading
+ */
 function lazyLode(view) {
-    //console.log('lazylode', view);
+	var container,
+		selector;
+		
 	switch (view) {
 		case 'radio':
-			setTimeout(function(){
-				$('img.lazy-radioview').lazyload({
-				    container: $('#radiocovers')
-				});
-			}, LAZYLOAD_TIMEOUT);
+			selector = 'img.lazy-radioview';
+			container = '#radiocovers';
 			break;
 		case 'tag':
-            if (SESSION.json['library_tagview_covers'] == 'No') {return;}
-			setTimeout(function(){
-				$('img.lazy-tagview').lazyload({
-				    container: $('#lib-album')
-				});
-			}, LAZYLOAD_TIMEOUT);
+            if (SESSION.json['library_tagview_covers'] == 'No') {break;}
+			selector = 'img.lazy-tagview';
+			container = '#lib-albums';
 			break;
 		case 'album':
+			selector = 'img.lazy-albumview';
+			container = '#lib-albumcover';
+			break;
+	}
+
+	if( selector && container) {
+		if ('loading' in HTMLImageElement.prototype) {
+			const images = document.querySelectorAll(selector);
+			images.forEach(img => {
+				img.src = img.dataset.original; // set the url back to the src				
+			});			
+		} else {		
 			setTimeout(function(){
-				$('img.lazy-albumview').lazyload({
-				    container: $('#lib-albumcover')
+				$(selector).lazyload({
+					container: $(container)
 				});
 			}, LAZYLOAD_TIMEOUT);
-			break;
 		}
+    }
+
 }
 
 function setFontSize() {
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -3231,11 +3231,11 @@ function lazyLode(view) {
 
 	if( selector && container) {
 		if ('loading' in HTMLImageElement.prototype) {
-			const images = document.querySelectorAll(selector);
-			images.forEach(img => {
-				img.src = img.dataset.original; // set the url back to the src				
-			});			
-		} else {		
+			var images = document.querySelectorAll(selector);
+			images.forEach( function ( img ) {
+				img.src = img.dataset.original; // set the url back to the src
+			});
+		} else {
 			setTimeout(function(){
 				$(selector).lazyload({
 					container: $(container)
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -1264,7 +1264,8 @@ function formatBrowseData(data, path, i, panel) {
 			if (data[i].file.substr(0, 5) == 'RADIO') {
 				if (panel == 'radio_panel') {
 					var imgurl = '../images/radio-logos/thumbs/' + filename.replace(path + '/', '') + '.jpg';
-					output += '"><div class="db-icon db-song db-browse db-action"><img class="lazy-radioview" loading="lazy" data-original="' + imgurl  + '"><div class="cover-menu" data-toggle="context" data-target="#context-menu-radio-item"></div></div><div class="db-entry db-song db-browse">';
+					var imgAttribute = isLazyLoadNative() ? 'src' : 'data-original';
+					output += '"><div class="db-icon db-song db-browse db-action"><img class="lazy-radioview" loading="lazy" '+imgAttribute+'="' + imgurl  + '"><div class="cover-menu" data-toggle="context" data-target="#context-menu-radio-item"></div></div><div class="db-entry db-song db-browse">';
 				}
 				else {
 					output += '"><div class="db-icon db-song db-browse db-action"><a class="btn" href="#notarget" data-toggle="context" data-target="#context-menu-radio-item"><i class="fas fa-microphone sx db-browse db-browse-icon"></i></a></div><div class="db-entry db-song db-browse">';
@@ -3201,6 +3202,11 @@ function setLibMenuHeader () {
 	$('#menu-header').text(headerText);
 }
 
+function isLazyLoadNative() {
+	//return false;
+	return 'loading' in HTMLImageElement.prototype;
+}
+
 /**
  * Enable lazy loading of images for a certain view.
  * Instead of src the image should have a data-original.
@@ -3210,40 +3216,34 @@ function setLibMenuHeader () {
  * @param {string} view - the view for lazy loading
  */
 function lazyLode(view) {
-	var container,
+	if( isLazyLoadNative() ==false ) {
+		var container,
 		selector;
-		
-	switch (view) {
-		case 'radio':
-			selector = 'img.lazy-radioview';
-			container = '#radiocovers';
-			break;
-		case 'tag':
-            if (SESSION.json['library_tagview_covers'] == 'No') {break;}
-			selector = 'img.lazy-tagview';
-			container = '#lib-albums';
-			break;
-		case 'album':
-			selector = 'img.lazy-albumview';
-			container = '#lib-albumcover';
-			break;
-	}
 
-	if( selector && container) {
-		if ('loading' in HTMLImageElement.prototype) {
-			var images = document.querySelectorAll(selector);
-			images.forEach( function ( img ) {
-				img.src = img.dataset.original; // set the url back to the src
-			});
-		} else {
+		switch (view) {
+			case 'radio':
+				selector = 'img.lazy-radioview';
+				container = '#radiocovers';
+				break;
+			case 'tag':
+				if (SESSION.json['library_tagview_covers'] == 'No') {break;}
+				selector = 'img.lazy-tagview';
+				container = '#lib-albums';
+				break;
+			case 'album':
+				selector = 'img.lazy-albumview';
+				container = '#lib-albumcover';
+				break;
+		}
+
+		if( selector && container) {
 			setTimeout(function(){
 				$(selector).lazyload({
 					container: $(container)
 				});
 			}, LAZYLOAD_TIMEOUT);
 		}
-    }
-
+	}
 }
 
 function setFontSize() {
--- a/www/js/scripts-library.js
+++ b/www/js/scripts-library.js
@@ -531,6 +531,7 @@ var renderAlbums = function() {
     // SESSION.json['library_encoded_at']
     // 0 = No (searchable), 1 = HD only, 2 = Text, 3 = Badge, 9 = No
     var encodedAtOption = parseInt(SESSION.json['library_encoded_at']);
+    var imgAttribute = isLazyLoadNative() ? 'src' : 'data-original';
 
 	for (var i = 0; i < filteredAlbums.length; i++) {
         filteredAlbums[i].year ? tagViewYear = '(' + filteredAlbums[i].year + ')' : tagViewYear = '';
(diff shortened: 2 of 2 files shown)


moodeaudio:
The construct below in the lazyLode() function causes error in minifier https://refresh-sf.com

```javascript
			images.forEach(img => {
				img.src = img.dataset.original; // set the url back to the src
			});
```

```
{
  "message": "Unexpected token: operator (>)",
  "filename": 0,
  "line": 3252,
  "col": 23,
  "pos": 120280
}
```

I can put function in the playerlib.nomin.js file to test but if u are using different minifier let me know.

moodeaudio:
Test results: 

Chrome uses native lazy loading and its fast. Safari does the fall back to jquery lazy load and its also fast. 

In the native lazy load case there's prolly a penalty for having to iterate with the forEach() ?

monkeypox:
Tim, 

I know iOS has an advanced pref to enable native lazyload in system prefs, not sure about MacOS as I’m not on catalina and the current safari tech preview for mojave is the wrong link.

It’d be better to test this once and then use the correct syntax while constructing the lists in scripts-library rather than doing it as in the demo but I assume it’s just being done that way because it’s a demo?

Anyway that would moot the minifier issue with the arrow function but if moode still wants to support older tablets (a good idea imo) then we need to be careful about arrow functions as they may not be supported.

moodeaudio:
I agree re constructing the lists. That could be the real PR at some point.

In any case good demo. I didn't even know the feature was being rolled into Browsers

bitkeeper:
Haha got the point; removed the es6 and ecma 2015 construction.

By the already very many changes for the upcoming release I can completely understand if you put this item in the fridge.

moodeaudio:
it works fine but as @monkeypox pointed out, for production it would be best to construct the lists up-front with the correct syntax rather than reconstruct them in lazyLode(). Maybe test for the capability at the top of panels.js, set a global flag and then test the flag to make the lists in scripts-library.js. Something like that.

monkeypox:
Right. I think it’s a good optimization, the older version of the lazyload library we have to use (because newer versions require intersectional observer) is buggy and annoying. It’d just be better to construct the lists using one method or the other depending on whether the browser supported native lazyload rather than having to adjust them afterwards.

monkeypox:
This is what I meant about doing it in the list construction, there's probably a better way to do this that doesn't require two strings but q&d:

In the renderAlbums function in scripts-library

With the declarations add:

	var imgstr1, imgstr2 = '';

then after it sorts the year stuff...

		if ('loading' in HTMLImageElement.prototype) {
			imgstr1, imgstr2 = '';
		} else {
			imgstr1 = '';	
			imgstr2 = '';	
		}
			if (SESSION.json['library_tagview_covers'] == 'Yes') {
			output += '<li class="lib-entry">'
             + imgstr
             + '<div class="tag-cover-text"><span class="album-name-art">' + filteredAlbums[i].album
             + '</span><span class="artist-name-art">' + filteredAlbums[i].artist + '</span><span class="album-year">' + tagViewYear + '</span></div></li>'		}
		else {
			output += '<li class="lib-entry"><span class="album-name">' + filteredAlbums[i].album
             + '</span><span class="artist-name">' + filteredAlbums[i].artist + tagViewYear + '</span></li>'
        }

		output2 += '<li class="lib-entry">'
            + imgstr2
            + '<div class="cover-menu" data-toggle="context" data-target="#context-menu-lib-all"></div>'
            + '<span class="album-name">' + filteredAlbumCovers[i].album + '</span>'
            + '<div class="artyear"><span class="artist-name">' + filteredAlbumCovers[i].artist + '</span><span class="album-year">' + albumViewYear + '</span></div></li>';
	}

bitkeeper:
@monkeypox while you was typing the message, I almost working on same idea to lose the query, PR is already updated withit.  

Maybe I should remove the word DEMO now ?

I instrumented the `lazyLode  `function 

original with scripted:
```
startup
scripted lazy for album takes 0.39499998092651367 ms
scripted lazy for album takes 77.06499996129423 ms

switch to first time radio
scripted lazy for radio takes 43.21000003255904 ms

switch to first time tag
scripted lazy for tag takes 40.36500002257526 ms
```

native lazy load boiler code (set attribute)
```
startup
native lazy for album takes 0.3349999897181988 ms
native lazy for album takes 95.72500002104789 ms

switch to first time radio
native lazy for radio takes 40.634999983012676 ms
switch to firt time tag
native lazy for tag takes 40.76000000350177 ms
```

after that manual calls from the console
```
native lazy for album takes 14.60999995470047 ms
native lazy for radio takes 13.52999999653548 ms
native lazy for tag takes 16.549999942071736 ms
```
native lazy load after get rid of the query (almost empty lazyLode  ) :
```
native lazy for album takes 0.0849999487400055 ms
```

moodeaudio:
I'll merge and check it out :-)

monkeypox:
Cool.

#1 It’d be better to set a global in playerlib as Tim suggested, then test against that rather than use a function call.

#2 I wish we could get rid of the unneeded classes if using native lazy loading but can’t think of a great way to do it. The way you implemented the check outside the loop will have the best performance, the best alternative to save memory would probably be to use a conditional and duplicate the output statements with the differing syntax. Personally unless we know memory is an issue I’d rather it be faster.

moodeaudio:
Right, after examining the code we should prolly do the following for least footprint / best performance.

playerlib.js

1. Leave existing lazyLode function as-is since it needs to support current way of doing lazy loading. It will become the fall back method as more and more browsers integrate native lazy.
2. Add "isLazyLoadNative" element to the GLOBAL array with default value = false.
3. In formatBrowseData() test GLOBAL.isLazyLoadNative instead of calling the function isLazyLoadNative().

panels.js

4. Test for native lazy load capability near top of code, maybe under the "// compile the ignore articles regEx" conditional.

```javascript
if ('loading' in HTMLImageElement.prototype) {
    GLOBAL.isLazyLoadNative = true;
}
```
scripts-library.js

5. In renderAlbums() test GLOBAL.isLazyLoadNative instead of call the function isLazyLoadNative().

@bitkeeper - let me know if this is acceptable or needs changes.

monkeypox:
One other thing we’ll probably want to do is more gracefully deal with missing artwork. There’s a way to do it with onerror but it seems to have a noticeable perf impact. Maybe in the thumbnail generator we can save a default thumb if there’s no artwork?

moodeaudio:
Thumb get already makes a moOde default thumb image if no cover found so I think we are "covered", lol

monkeypox:
> playerlib.js
> 
> Leave existing lazyLode function as-is since it needs to support current way of doing lazy loading. It will become the fall back method as more and more browsers integrate native lazy.
This needs to change to something like the version Marcel submitted in order to bypass lazylode for native. After that it’s basically the same thing but moves the timeout to the end.
> Add "isLazyLoadNative" element to the GLOBAL array with default value = false.
It shouldn’t be a question imo, just GLOBAL.nativeLazyLoad maybe?

monkeypox:
> Thumb generator already makes a moOde default thumb image if no cover found so I think we are "covered", lol

Displaying the default image on error was something I added to the lazyload library, this is what happens with native lazy load which I don’t mind but it looks awful in Chrome.

9223B44F-CDB4-430F-B127-31C70A49BD69

moodeaudio:
Right, Marcel's mod to lazyLode() is in fact needed. 
Ah, the mods for default gray bg and the default image.

Maybe there is a clever way to do default image in native mode but if not then its only a going to be a visual issue for those collections that are not well tagged. If we can do something we will but otherwise I don't see this as a reason not to implement given the efficiency of native.

bitkeeper:
Maybe leave `isLazyLoadNative` in place. The call will cost you about 0.01 ms, so it isn't a performance issue.

In case of problems with certain browser, the function is an easy extension place to force lazy or no native lazy support for a certain platform or browser.

@monkeypox I was not sure if other class could be remove; due use in styling orso ?

Concerning the artwork; the big album covers are already loaded by coverart.php which supplies an default in case of missing artwork. To be save we could load the thumbnails also by using the same construction.

__OFFTOPIC ON__
While refactoring parts of the frontend for ncreased build tooling support like gulp. I was also thinking why not move the readio artwork outside the code try (/var/www) and move it to /var/local/www.
Instead of additional nginx location (what already is an improvement), why not let is also be provide by an php helper. Maybe even the same coverart.php
__OFFTOPIC OFF__

moodeaudio:
Radio covers and thumbs are fetched based on the .pls base name and not whether they have a folger.jpg or embedded cover in a song file which is what coverart.php handles.

For example to fetch images for Soma FM - Groove Salad.pls the code just does Soma FM - Groove Salad.jpg directly. No need to search. There are exceptions but thats encoded in the logo column in cfg_radio table.

The radio images could be moved to /var/local/www. IIRC they were there in the past when we were using read-only squashfs for /var/www.

monkeypox:
@Marcel
> Maybe leave isLazyLoadNative in place. The call will cost you about 0.01 ms, so it isn't a performance issue.
> 
It’s not performance in this case but more about how we want boolean type flags to work. If it’s set in GLOBAL on load then it’s also easy to test the opposite case even on a browser that supports native lazy loading. 
> In case of problems with certain browser, the function is an easy extension place to force lazy or no native lazy support for a certain platform or browser.
> 
I dont think it’s a problem with any browser, we just need to sort how/if the default image is handled.
> @monkeypox I was not sure if other class could be remove; due use in styling orso ?
> 
Those are used exclusively by the lazyload library to determine if they should be included, the use of different class names instead of just “lazy” was to work around a bug in the library. The only real downside to including the unused syntax is memory consumption. If it turns out to be an issue like with really large libraries we can revisit it otherwise it’s mostly an OCD type thing. 
> Concerning the artwork; the big album covers are already loaded by coverart.php which supplies an default in case of missing artwork. To be save we could load the thumbnails also by using the same construction.
> 
That’s likely to introduce significant overhead.

It’s easy to add something like:

onerror=“this.src = defaultcover.jpg”

to the img tag but it seemed to noticeably slow things down. 

Theoretically you can do a global event handler like:

$(“img”).on(“error”, function(){
    $(this).attr(“src”, UI.defCover);
});

But I can’t get it to work for this, maybe because of lazy loading. I’ll search around a bit.

If we could save a default image to the thumb cache for albums without artwork that’d only cost us a little disk space. The downside is the current method uses an .svg file so using a jpg we’d lose some text sharpness and it won’t be translucent, the upside is it’d be faster.

monkeypox:
I looked into this and it turns out thumbcache.php is already set up to do the substitute file thing I was talking about, we just need to add a file named “nothingfound.jpg” to /var/www/images (and rebuild the thumb cache). I made a q&d 400x400 example:

nothingfound

monkeypox:
Expanding on Marcel's idea a bit we can remove the extra syntax by doing this outside the loop:

	if (GLOBAL.nativelazy) {
		var imgstr1, imgstr2 = ''
         	+ imgstr1 + filteredAlbums[i].imgurl + '">'
			+ '<div class="tag-cover-text"><span class="album-name-art">' + filteredAlbums[i].album
            + '</span><span class="artist-name-art">' + filteredAlbums[i].artist + '</span><span class="album-year">' + tagViewYear + '</span></div></li>'		}
		else {
			output += '<li class="lib-entry"><span class="album-name">' + filteredAlbums[i].album
            + '</span><span class="artist-name">' + filteredAlbums[i].artist + tagViewYear + '</span></li>'
        }
		output2 += '<li class="lib-entry">'
            + imgstr2 + filteredAlbumCovers[i].imgurl + '">'
		+ '<div class="cover-menu" data-toggle="context" data-target="#context-menu-lib-all"></div>'
		+ '<span class="album-name">' + filteredAlbumCovers[i].album + '</span>'
		+ '<div class="artyear"><span class="artist-name">' + filteredAlbumCovers[i].artist + '</span><span class="album-year">' + albumViewYear + '</span></div></li>';

After some testing it seems like native lazy loading is slower on iOS (there's a noticeable delay before it starts displaying images, probably why it's in experimental features). I wouldn't recommend enabling this right now except for testing.

There's an optimization Chrome mentions in the console that we can enable by setting h/w in the tag but we'd need to calculate a pixel value like...

	var hwc = ($(window).width() * parseInt(getComputedStyle(document.documentElement).getPropertyValue('--thumbcols')) / 100) * .9;

...and then add that value as h/w to the native lazy imgstr2 like...

		var imgstr2 = '<img loading="lazy" height="' + hwc + 'px" width="' + hwc + 'px" src="';

...and remove the h/w values from the css as the image is already sized. Something similar would also need to be done with imgstr1 for the tag images.

bitkeeper:
My 2 cents;I personally prefer to keep it in the loop, the performance loss is near nothing. Espcially if you see what other things are already happening in the lines above it. While the clarity of the code is better, because the code isn't scattert arround.

Concerning the problems with the safari browser, there will be always be browsers that will not react the way as espected.

May add add useragnet to the equation(opt in or opt out) like:
```
'loading' in HTMLImageElement.prototype) &&
(navigator.userAgent.indexOf("Firefox") != -1 || 
navigator.userAgent.indexOf("Chrome") != -1)
```
or
```
'loading' in HTMLImageElement.prototype) &&
navigator.userAgent.indexOf("Safari") == -1  
```

monkeypox:
You already have src and data-original set outside the loop, the suggestion is basically the same thing except it removes the superfluous loading / classes code situation in the pr. I don’t see any benefit to wasting client memory or putting a conditional inside the loop to be executed thousands of times when you don’t need to do it. The year code unfortunately really does have to be inside the loop. Oh. I’m a few versions back but looking at the current code in develop all the encoded at stuff should be tucked inside a conditional.

Safari doesn’t support native lazy loading by default so no need to build in a lock out that isn’t needed now and may never be.

moodeaudio:
@monkeypox 
- Can you clarify the issue you are seeing with with encoded_at? It's a fancy option and has to be tested for each album.
- I haven't looked at thmcache.php in quite a while but I'll check it out later today.

monkeypox:
I meant that it appears to have an option to not show the info (9) so it should have a conditional to bypass the code if it’s not going to be used.

If you define the vars outside the loop as null with the rest of the declarations you can avoid declaring them (e.g. var foo) inside the loop and as the code that sets them would be skipped they’ll just be empty strings and not affect the layout.

moodeaudio:
Ah, now I see the faulty logic :-0

moodeaudio:
We can prolly do something like below which handles the No case.

```javascript
    // SESSION.json['library_encoded_at']
    // 0 = No (searchable), 1 = HD only, 2 = Text, 3 = Badge, 9 = No
    var encodedAtOption = parseInt(SESSION.json['library_encoded_at']);
    var tagViewHdDiv, tagViewNvDiv, albumViewNvDiv, albumViewHdDiv, albumViewTxDiv, albumViewBgDiv = '';

	for (var i = 0; i < filteredAlbums.length; i++) {
        filteredAlbums[i].year ? tagViewYear = '(' + filteredAlbums[i].year + ')' : tagViewYear = '';
        filteredAlbumCovers[i].year ? albumViewYear = '(' + filteredAlbumCovers[i].year + ')' : albumViewYear = '';

        // filteredAlbums[i].encoded_at
        // [0] bits/rate format. [1] flag: "l" lossy, "s" standard def or "h" high def
        if (encodedAtOption != 9) {
            // Tag view
            var tagViewHdDiv = encodedAtOption == 1 && filteredAlbums[i].encoded_at.split(',')[1] == 'h' ? '<div class="encoded-at-hdonly-tagview">HD</div>' : '';
            var tagViewNvDiv = encodedAtOption <= 1 ? '<div class="encoded-at-notvisible">' + filteredAlbums[i].encoded_at.split(',')[0] + '</div>' : '';
            // Album view
            var encodedAt = filteredAlbumCovers[i].encoded_at.split(',');
            var albumViewNvDiv = encodedAtOption <= 1 ? '<div class="encoded-at-notvisible">' + filteredAlbumCovers[i].encoded_at.split(',')[0] + '</div>' : '';
            var albumViewHdDiv = encodedAtOption == 1 && encodedAt[1] == 'h' ? '<div class="encoded-at-hdonly">HD</div>' : '';
            var albumViewTxDiv = encodedAtOption == 2 ? '<div class="encoded-at-text">' + encodedAt[0] + '</div>' : '';
            var albumViewBgDiv = encodedAtOption == 3 ? '<div class="encoded-at-badge">' + encodedAt[0] + '</div>' : '';
        }
```

moodeaudio:
Re earlier post, the thmcache generator needs a bit of work cos there is some baggage logic from coverart.php which was the basis for thmcache. 

Yes, we could also create a notfound.jpg file to compensate for native lazy load "no image" situation.

Given that native lazy load is not yet universal and we have a decent Jquery lazy load, what is compelling us to add the conditionals and extra code for native?

monkeypox:
Tim,
> Re earlier post, the thmcache generator needs a bit of work cos there is some baggage logic from coverart.php which was the basis for thmcache.
> 
> Yes, we could also create a notfound.jpg file to compensate for native lazy load "no image" situation.
> 
Aside from missing the features of SVG I think it’s probably better to use the nothingfound.jpg feature to address missing artwork. Currently it generates a console error which is always annoying and seems to cause moode to stutter a little (possibly because svg is more expensive to display).
> Given that native lazy load is not yet universal and we have a decent Jquery lazy load, what is compelling us to add the conditionals and extra code for native?
> 
I think it’s just a matter of skating to where the puck is going to be kind of thing. Here’s a nice overview of things from a web developer perspective:

https://css-tricks.com/a-deep-dive-into-native-lazy-loading-for-images-and-frames/

Native is supported on Chrome / Blink and WebKit seems committed to implementing it even if it’s classed as experimental for now. With the h/w added to the img tag it does seem faster though that‘s possibly a general perf improvement from using a fixed pixel size that’s calculated once instead of using a more complex css rule for every img (the downside is it uses extra memory). With h/w added to the tag the native path skips a partial file request to sniff h/w (and in our case we don’t really use actual image h/w anyway). I didn’t see any benefit to having the first x number of items classed as loading=“eager”.

Anyway I think it’s definitely worth doing even if it doesn’t make the next point release.

moodeaudio:
Gotcha. 

Rather than update this PR I'll just go ahead and manually add the native lazy stuff since it's pretty straight forward and then I'll push a commit later this morning.

moodeaudio:
Since the capability has been implemented outside of this PR and heavily based on it there is no need to keep it open.

PR #146, opened 2020-07-05
Proposal to hide the volume controls when MPD volume is disabled

shikasta-net:
It has always bugged me that the volume controls are displayed when I have MPD volume disabled.  I tried a quick hack to see if I could remove it easily.  I'm happy to improve this.

Load the MPD config
If the `mixer_type` is disabled then set a flag (I've hacked in `style="display:none"` because I wanted to see if it would work before digging into the CSS properly)
else display as usual.

Looking forward to feedback.

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -27,6 +27,16 @@ require_once dirname(__FILE__) . '/inc/playerlib.php';
 
 playerSession('open', '', '');
 session_write_close();
+$dbh = cfgdb_connect();
+
+// Load settings
+$result = cfgdb_read('cfg_mpd', $dbh);
+$mpdconf = array();
+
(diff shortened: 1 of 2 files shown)


moodeaudio:
It's come up once or twice in the past but the problem is that without the volume knob occupying the space between the play controls and the toggle buttons, the symmetry of the UI layout becomes broken and It ends up looking very odd.

shikasta-net:
That makes sense.  To me (and others it sounds like), the erroneous control is more bothersome.  I will play about with the layout and see if there's a solution that could satisfy both.  I'll post screenshots.

moodeaudio:
Closed due to lack of activity.

PR #149, opened 2020-07-19
Move header.php from /var/local/www to /var/www.

bitkeeper:
Not any longer required to be in /var/local/www so move it to the place where the other sources are.


moodeaudio:
If we ever decide to deliver /var/www as a squashfs, which was the case in an earlier moOde series, then we can't include header.php in that dir cos the <title> text (Browser title) gets updated.

bitkeeper:
Maybe I am missing something, but can't we just fill in the title directly from $ _SESSION ?

Like:
```php
<title><?php echo $_SESSION['browsertitle']; ?></title>
```
Or is there important reason to change the code itself when the browser title is changed?

At least it works in my test environment and this construction will still work if you ever decide to use the squashfs again.
In that case no need to update the file and as benefit some code can be removed like the call to util.sh.

moodeaudio:
That should work. I'll merge this and then make the changes to use the session var.

bitkeeper:
@moodeaudio 

To be sure; also the sed command should be made comment in `command/util.sh` :

```php
	if [[ $2 = "browsertitle" ]]; then
		sed -i "s/<title>$3/<title>$4/" /var/local/www/header.php
	fi
```

Later on maybe some more cleaning todo, but that will take some more time.

moodeaudio:
Yes, there's a few more locations where the SED needs to be commented out.

I push the commits a bit later.

PR #161, opened 2020-07-28
show cover art at the playlist

bitkeeper:
Show (album) cover at the playlist for radio and tracks from the mpd database.


bitkeeper:
@monkeypox and @moodeaudio: I am missing something with enabling the non-native lazyload, can you help me out ?

And for the source that don't support cover art, what is desired behavior ?:

- Blank placeholder
- The moOde logo as for not found 
- ???

moodeaudio:
In playerlib.js at line 1000

```
		if( GLOBAL.nativeLazyLoad && option_show_playlistart) {
 			lazyLode('playlist');
 		}
```

should prolly be

```
		if( option_show_playlistart) {
 			lazyLode('playlist');
 		}
```
because the lazyLode function already includes the test for GLOBAL.nativeLazyLoad.

bitkeeper:
Haha thanks! It clearly too late tonight .

moodeaudio:
Trying to resolve the original conflict in the PR 's .sql file I deleted it and then added the new sql column manually and then uploaded a new .sql file to the repo.

This didn't work as I expected... If you wouldn't mind, I'll need some help to fix things up so I can merge the PR.

bitkeeper:
No proplemo, fixed the conflict and synced with the current state of the develop head.

steps done in local workingtree:
- pulled latested state of playlist_coverart
- reset the playlist_coverart just the moment before the delete of sql file.
- pull of remote develop 
- fixed the conflict of sql by accepting your version (already included the new record for playlist art)
- marked it as resolved with git commit
- used a push with --force to override the current state in repo (due the repo was 1 commit ahead due the delete sql file)

moodeaudio:
One side effect of showing the covers is that it reveals how often the playlist is rendered. The current design renders it every time renderUI() happens :-0 This means that the thumbs are reloaded unnecessarily.

I added a bit of test logic in that last commit of playerlib.js that serves to control when the playlist is rendered but I'm sure there is a better way. Maybe a redesign of playlist rendering/updaing?

bitkeeper:
Yes I also noticed that the playlist render is called more often than required.  Only I didn't dare to touch that piece of code, because I was not sure of the working and the side effects etc.

I see already commits by you to change this rendering to improve it. Great! That wil benefit everybody playlist thumbs or not.

After an initial load the images aren't really reloaded, but served from browser cache.

monkeypox:
This is vaguely familiar, iirc when working on making large playlists usable we discovered it redrawing the playlist unnecessarily (and with a big list that’s no bueno) but I thought we addressed it. Or at least I didn’t notice it when using larger playlists where it would show up, but I haven’t done that in a while as mpd .22 was crashing after a while with 10k+ tracks so maybe a regression crept in (or this is unrelated). Anyway there’s definitely potential to minimize work in the renderui function.

moodeaudio:
lol, some of that JS code is really scary isn't it :-0 Can never rule out a regression but I don't think in this case it's what we are seeing. I remember way back there was some sort of conditional that was controlling when to render the Playlist but it wasn't working right so I prolly just nuked it.

Even though the images are subsequently served from browser cache it's the visual disruption of seeing them re-populate the list thats not so good.

So far that last commit seems to be working ok. Lets live with it for a while and then see if there are some improvements that could be made.

Issue #173, opened 2020-08-12
Album art not displayed when playlist contains extended m3u directives


sputnik13:
When an m3u playlist is loaded with optional extended m3u directives that include the display title of the song (see https://en.wikipedia.org/wiki/M3U#Extended_M3U under #EXTINF), MPD returns the display title as Name in the currentsong command.

This breaks the assumption in https://github.com/moode-player/moode/blob/develop/www/inc/playerlib.php#L3186 where the presence of a Name field assumes that a radio station is being played.

This causes any playlist items loaded with #EXTINF populated to always return the default coverart.

moodeaudio:
Try a different logical grouping of the elements in the statement. For example.
```
// Radio station
elseif ((isset($song['Name']) || substr($song['file'], 0, 4) == 'http') && !isset($current['duration'])) {
```

moodeaudio:
My proposed logic change would not have worked. 

After testing, PR 174 accepted.

PR #174, opened 2020-08-12
remove Name check for radio station

sputnik13:
Remove check for Name field in mpd response when checking for Radio station as Name can be populated via extended m3u directives when a playlist is loaded.

Addresses #173

### DIFF
--- a/www/inc/playerlib.php
+++ b/www/inc/playerlib.php
@@ -3183,7 +3183,7 @@ function enhanceMetadata($current, $sock, $caller = '') {
 			//debugLog('enhanceMetadata(): iTunes AAC or AIFF file');
 		}
 		// Radio station
-		elseif (isset($song['Name']) || (substr($song['file'], 0, 4) == 'http' && /*!isset($song['Artist'])*/!isset($current['duration']))) {
+		elseif (substr($song['file'], 0, 4) == 'http' && /*!isset($song['Artist'])*/!isset($current['duration'])) {
 			debugLog('enhanceMetadata(): Radio station');
 			$current['artist'] = 'Radio station';
 


moodeaudio:
Please refrain from cross-posting! There is a Git issue, a PR and also a Moode Forum post all referencing the same question / issue. Next time, just choose one to get the ball rolling.

As far as removing the test for Name, that can't be done unless there are other tag(s) in MPD currentsong data that can be used to reliably identify a Radio station or equivalently some sort of change to the logic in playerlib.ohp enhanceMetadata function for identifying a Radio station.

I responded to your issue with a proposed logic change. Try it out and let me know how it goes (in the issue).

sputnik13:
@moodeaudio sorry, force of habit, I started in the forum then moved to github when I realized it's a code issue.  When you say try it in the issue, are you saying the project doesn't accept PRs?  Just want to understand what your expected process is as I didn't see a contributor's guide in the repo.

sputnik13:
also, Name is fairly optional, seems like checking for a http prefix in the file is the most reliable way to see whether the audio being played is streamed off a website as file is a required attribute for the currentsong command per MPD documentation https://www.musicpd.org/doc/html/protocol.html#querying-mpd-s-status

moodeaudio:
Song file URL's from UPnP AV media servers are also prefixed with http so that alone can't be used to identify Radio stations. 

The Name tag is normally part of Radio station metadata and the current test which uses Name has worked reliably for many years. Your proposal to remove Name requires that you provide a way of differentiating between Radio stations and UPnP song file URL's thats as reliable as the current method.

Of course PR's are accepted but you provided multiple vectors to respond and I chose the issue. You could close the issue and just use this PR to continue the discussion.

sputnik13:
Sorry, it's not my intention to create more work for you.  The lack of a contributor's guide led me to snap practices that have been common in other open source projects I've worked on where they typically want an Issue filed documenting the problem with a PR to identify and rev on the work.

I understand now this is not your preferred mode of communication so I will not go back to the other channels.  I think a contributor's guide would be really helpful for making clear what your preferred mode of communication and interaction is.

With respect to using Name for identifying Radio stations, I understand that Radio stations commonly use this field.  The point I'm making is that it's also true that any time Name is present where it would be a radio station, the file would have an http prefix.  So checking for the latter gets you the match you want without having to check for the former.  The caveat that you have to distinguish between UPnP and a radio station exists regardless of whether you check for Name or not as Name is a field that can also be present in m3u lists.

So for a radio station, unless there is ever a case where file would not have an http prefix, then the check for the Name field is extraneous.  The duration field check then clarifies that the absence of the duration field makes it highly likely that it's a radio station.

Seems like it's an opportunity to simplify the logic with no loss in fidelity for the match.

moodeaudio:
Our project is small and lacks some of the formality of other projects but it hasn't really been a problem. I'll work with whatever you send in :-)

I see your point about the logic and while I was examining the code it appears that a test for just http && no duration tag might suffice to identify radio stations. 

I'll do some testing on my end and report the results.

moodeaudio:
I didn't experience any side effects from the change. Thanks for the contrib :-)

PR #177, opened 2020-08-25
Updated moode-sqlite3.db.sql to add a new station

hippiecow:
Added the stream for The Classical Station, WCPE 89.7 FM out of Wake Forest, NC.


moodeaudio:
What would be needed for this to be considered is the .pls file, a good quality 400x400px or higher res station logo image and a 200x200px thumb.

You can also just email me the station stuff or post it in the moOde Forum and I'll check it out. I think that would be easier than a PR.

hippiecow:
Thanks, Tim. I’ve attached a zip containing a .pls and two images. 

Best regards,
Dean

moodeaudio:
No attachments are present.

PR #181, opened 2020-09-11
ignore que files in music lib

bitkeeper:
Add a default .mpdignore to ignore *.cue files.

### DIFF
--- a/other/build/build_recipe_v2.26.txt
+++ b/other/build/build_recipe_v2.26.txt
@@ -338,6 +338,7 @@ sudo chown -R mpd:audio /var/log/mpd
 sudo cp ./moode/mpd/mpd.conf.default /etc/mpd.conf
 sudo chown mpd:audio /etc/mpd.conf
 sudo chmod 0666 /etc/mpd.conf
+sudo bash -c 'echo -e "**/*.cue\n**/*.CUE" >/var/lib/mpd/music/.mpdignore'
 
 2. Install MPD dev libs.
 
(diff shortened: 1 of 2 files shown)


moodeaudio:
Prolly a good idea but it would need to be an option in the Library section of Appearance because some users do want their CUE sheets indexed by MPD. This could be part of the post 7.0.0 feature set and be generalized to allow user to enter the ignore string.

bitkeeper:
Haha I already expected that you gonna say that.

Exactly "some user" require que files, most don't and I see it in the forum often popup as a support question with 'double' item in the library. This behaviour is introduced some time agone with a certain MPD version. Prop in the past the default settting where different.
So I would say add default the .mpdignore with **/*.cue and let the minority of the users change this file (at least it exists then) to include que files (especially because this behaviour changed 2 mpd versions back or so).

And fix it to be configuratable in post 7.0.0.

Anyway at least we now have a door hanger as reminder ;-)

richo67:
Hello, I hope you do not mind me to add a view on this here.
I am one who like to have certain cue visible, but not all of them. 
Not sure if this is generic situation, but cue sheet has its reason to be indexed only if all tracks in it points to one huge file (often flac in my case). This is useful when there is no space between tracks. In this case the flac file indexed this way bye cue should not be visible as one track file (whole album).
If however cue sheet is pointing for each track to different flac then it could be either ignored (metadata are taken from the music file itself) or the information in cue should complement missing data from the music files.

bitkeeper:
@richo67 thanks for your reaction.
Don't worry, I understand the usefulness of que files. It did only concerns the default content of mpd .mpdignore file, which is default absent with moOde. As @moodeaudio states if supported it will be configuratable implementation.

PR #188, opened 2020-09-16
Extending moodecfg.txt feature

TookaFace:
- Changing sections name for matching the names in web UI
 - Adding all the renderers and the possibility to resuming
 - Adding ability to set I2S device
 - Adding system options: keyboard, local UI, CPU Governor and Wait for eth0 address
 - Adding more theme options
 - Removing mpdzeroconf option, since we do not have the possibility to change it in web UI
 - Updating setup.txt and moodecfg.txt

I've running tests, that seems to work, you can see a autocfg.log from one of my test
[autocfg.log](https://github.com/moode-player/moode/files/5233647/autocfg.log)

### DIFF
--- a/boot/moodecfg.txt.default
+++ b/boot/moodecfg.txt.default
@@ -65,4 +65,4 @@ cover_scale=1.25
 [other]
 font_size=Normal
 playhist=No
-first_use_help=y,y
+first_use_help=Yes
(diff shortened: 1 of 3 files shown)


TookaFace:
Depending on what you want for this, i could expand a lot more, i already have a python script who configure almost entirely moOde, so it's easy for me to transcript in the autoConfig function.

PR #189, opened 2020-09-17
set correct exec flag of the files in www

bitkeeper:
some scripts where missing an +x flag and othe files have a +x flag but aren't executable like images. Checked the www and subdirs.


moodeaudio:
I'm not sure I understand this PR. Does it actually change any of the files?

bitkeeper:
A lot of non executable files like images and javascript have the executable flag set.
And some scripts in /var/www/command are missing the executables.

This PR change the excutable flags(chmod -x or chmod +x)  of the files, not content changes.

bitkeeper:
Hold it !
See something strange in the long list of changes ... some files are delete not sure why, but let me fix this first.

bitkeeper:
Ok fixed.

moodeaudio:
What does the PR actually change? When I look at the "Files changed" list every file has "No changes" under it.

bitkeeper:
No content update, only file attributes:
image

is equivalent to:
chmod +x www/command/sysinfo.php
chmod -x www/css/bootstrap.css

moodeaudio:
Got it :-)

moodeaudio:
I just merged this (GitHub Desktop) and my changes list showed a bunch of filed deleted :-0 Then when I looked here the PR was still un-merged and it had a list of conflicts which appeared to be the list of files marked as deleted.

I reverted the merge and pushed the commit but I'm really confused as to what happened ?

GitHub and I are just not getting along today...

bitkeeper:
It looks like the repo state was used from before I fixed it, because that one did contain the deletes of file in.
Can you try again, but now from the github repo interface (here) ?

btw if the last commit to a branch is unwanted instead of a revert you can also reset the head one step back, this prevent that you see merge followe by reverted.

moodeaudio:
I noticed that in GitHub Desktop the PR listed 448 changed files as compared to whats shown here which is 202 changed files.

bitkeeper:
Prob I caused that by undoing the changes 448 files, redoing to 202 followed by a push with --force.
As long as nobody is using the branch that is ok, but the moment the branch was already used (Github Desktop) it requires that a pull with also --force is done.

moodeaudio:
Here's a screenie after switching to the PR 189 branch. Should I pull these commits and would that get the PR current with whats in the Git repo?

Screen Shot 2020-09-19 at 2 33 27 PM

bitkeeper:
Can you indicate that the pull should be forced else do it from the webinterface ?

moodeaudio:
There is no option in Git Desktop for forcing the pull so I'll do the merge from here.

moodeaudio:
After merging here, Git Desktop prompted me to pull down two commits from the repro which I did.

Looks like we are all sync'd up now :-)

bitkeeper:
Great, next time I'm doing a forced push I will warn about it.

PR #190, opened 2020-09-20
Add height/width to thumbnail image element

monkeypox:
This adds height & width to the thumbnail image element which is supposedly a perf boost for native lazy load but also solves the issue with how images appear while scrolling up in a thumb view.

This also adds back some code that got lost in the ill-fated tidal artwork "fix" pr, will re-open if I can fix conflicts...


monkeypox:
deleted this cursed branch and moved to a new one...

moodeaudio:
Maybe there are other branches that can be deleted?

PR #192, opened 2020-09-23
Expandalbum

scripple:
Create behavior that clicking a selected album which is only showing the tracks from a non-album artist on a compilation album will expand the tracks to all tracks on the album.


moodeaudio:
The toggle part of the feature does not seem to work for me.

Repro:

- Click on artist that belongs to a comp album
- the comp album is listed
- the tracks belonging to the artist are listed

- click the album
- all tracks in the album are listed

- click the album again
- all tracks in the album are listed (no change)

Expected behavior

- only the tracks belonging to the artist are listed

scripple:
I didn't realize a compact feature was necessary.  (Click the artist again or click another album and back works.)  There's no actual state kept to say it has expanded.  So it would either need a state variable, or it would have to filter songs twice and compare which list isn't active and use that.  Can be done though.  Have a preference?

scripple:
The other filters aren't toggles either.

moodeaudio:
Ok I see that the toggle is dismissed by tapping on the artist.

Nvm

moodeaudio:
It will be a challenge to make the help for this noice feature

scripple:
Agreed.  Fortunately nobody will read the documentation anyway unless it's a popup on the albums.  I only found out looking through the code that the tops of the columns were reset buttons for the filters in those columns.  I'm sure you've documented it somewhere though.

It would be a shame to lose it just because of the documentation issues though.

moodeaudio:
Why can't you make the documentation for your new feature?

scripple:
I don't even know where the documentation for moode lives.  We could just leave it an easter egg.

Issue #197, opened 2020-09-24
More robuste autocfg handling idea


bitkeeper:
@moodeaudio and @TookaFace about making the autocfg implementation a little bit more robust:
Adding `if` construction arround each config setting handling would explode the code. Also for each config setting also a logprint is already present.

How do you guys feel about making a more generic implementation ?
With a configuration how autocfg setting should be handled.
Generic loop can than be used to handle the checking logging and calling the actual handler.

To get the idea a proto is below:

```php
// Auto-configure settings at worker startup
function autoConfig($cfgfile) {
	...
	// above existing filling of $autocfg
	
	// Define configurationHandlers with lambda functions, 
        //   - cfgkeys define which autocfg entries are required for this block
        //  - section name is only used to print the current section
       //   - handler function to handle the configuration changes, as argument a dictionary
	$configurationHandlers = [
		array( 'cfgkeys' => ['hostname'] ,'section' => 'names', 'handler' => function($values) {
			playerSession('write', 'hostname', $value['hostname']);
		}),
		array( 'cfgkeys' => ['wlanssid', 'wlanpwd'] ,'section' => 'network', 'handler' => function($values) {
			// your implementation
		} )
	];

	// shorthand to dictionary keys
        $available_configs = array_keys($autocfg);
	$section ='';
        // central handling and checking loop for autocfg settings
	foreach ($configurationHandlers as $config) {
		$values = array();

		// print new section header
		if($section != $config['section']) {
			$section = $config['section'];
			autoCfgLog('autocfg: - '. $section);
		}

		// check if alrequired cfgkeys are present
		if( !array_diff_key(array_flip($config['cfgkeys']), $available_configs) ) {
			// is so get copie all key/value sets
			foreach ($config['cfgkeys'] as $config_name) {
				$value = $autocfg[$config_name];
				$values[$config_name] = $value;
				autoCfgLog('autocfg: '. $config_name.': ' . $value);
				// remove used autoconfig
				unset($available_configs[$config_name]);
			}
			// call handler
			$config['handler'] ($values);
		}
	}

	// check for unused but supplied autocfg settings
	if( empty($available_configs) ) {
		foreach ($available_configs as $config_name) {
			autoCfgLog('autocfg: '. $config_name.': unused configuration setting, incomplete or wrong name.');
		}
	}

```

moodeaudio:
Converting from all that in-line code to a compact loop would be great.

There will need to be a switch case to handle the configs that involve more than just updating a SQL column.

[...]

PR #199, opened 2020-09-25
Refactored autocfg import and added support for generation of the autocfg

bitkeeper:
This is a proto; not tested just for the dicussion at #197.
Don't look at the diff, but a the whole function autoConfig.


bitkeeper:
@moodeaudio  I think we can merge it.
@TookaFace If more config options should be support this can be done in the function `autoConfigSettings` by adding an item to  the $configurationHandlers array

Most simple form of such item:
```
['requires' => ['airplayname'] , 'handler' => setPlayerSession],
```
This that it requires the config option 'airplayname' to be present. No custom handler is need, just use the standard `setPlayerSession` handler.

Other example that use playsession and /var/www/command/util.sh:
```
['requires' => ['hostname'] , 'handler' => setPlayerSessionAndSysCmd, 'cmd' => 'chg-name host "moode" "%s"'],
```
Here the standard handler `setPlayerSessionAndSysCmd` is used which also supports a the supplied cmd to be executed with util.sh.

Also custom handlers for import (handler) and or export (custom_write) is supported.
More examples are in $configurationHandlers.

TookaFace:
Ok, it seems straightforward, i will give a look once commit

PR #201, opened 2020-09-25
Make clicking an expanded album toggle back to showing just the selected artist.

scripple:
Here's your toggle.  State stored in css like active.


scripple:
I thought of a way that doesn't require state and thus doesn't affect any other part of the code.  Was concerned calling removeClass on every album for a second class even in a single call could be slow for a large album collection.

Tested with a single artist, multiple artists ctrl-selected including some that shared the same album and different albums, and also a test including the compilation album artist.  All seemed to work ok.

PR #205, opened 2020-09-28
WIP for addressing library lazy load issues

monkeypox:
The lazyload library was never designed to deal with dynamically added content so we used settimeout statements to pause for the lists to populate. This doesn't always happen in time which results in no thumbs.

This implements .when().then() to the list output routines with the then() being a call to lazyLode(). This lets the list finish populating before lazyload() is ever called. All other subsequent calls to .lazyload() shouldn't need to be placed in a settimeout statement.

This also:

#1 limits the scope of the .limited class some more

#2 fixes a playlist thumb margin issue

#3 hides the e.g. radio search reset by default so it doesn't flash on/off on page reload

#4 adds object-fit as discussed on the forum

#5 adds lazylode calls to the thumb column setting as when it increases more thumbs are exposed, etc.

#6 reverts the conditional in the setinterval to use coverview & currentView

#7 fixes radio search (span div swap)

### DIFF
--- a/www/css/panels.css
+++ b/www/css/panels.css
@@ -210,7 +210,7 @@ img.coverart {border-style:none;border-radius:.1em;box-shadow:0 .1em .2em rgba(3
 #albumsList li {padding-top:.4em;padding-bottom:.4em;}
 #albumsList .tag-cover-text {display:inline-block;vertical-align:middle;transform:translate(0, .2em);min-width:50%;max-width:calc(100% - 3.6rem);/*line-height:1.6rem;*/}
 #library-panel .lib-entry span {max-width:100%;display:inline;}
-#library-panel.limited .lib-entry span {display:inline-block;}
+#library-panel.limited .album-name, #library-panel.limited .artist-name, #library-panel.limited .album-year {display:inline-block;}
 #top-columns .lib-entry {font-size: 1em;}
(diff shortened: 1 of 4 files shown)


moodeaudio:
Theres a couple of conflicts. I assume u want all the swizzle's applied and none from develop, correct?

monkeypox:
Hi Tim,

In panels yes.

In playerlib do the first one swizzle



the second one only the library_thumbnails line from swizzle, it’s just tidying it to be like the other lines.



the third you can keep either



though default_timeout seems fine with my testing. I have radio & playlist do like:

$(selector).html(output, lazyLode(‘radio’))

This is supposed to let the lazylode execute after the output() completes, possibly. Well... it seems to, but it worries me that it doesn’t work with the library lists so I’m going to try another method later and see how that goes.

moodeaudio:
I tried the panels changes manually and ran into some breakage in the Library cos of the need to have the CSS applied to the spans which are generated.

I think this PR should prolly be broken into 3 smaller PR's that each focus on a specific area. It would make it a lot easier to test.

monkeypox:
The last commit should have reverted that and fixed the genres & artists columns wrt applying text-overflow:ellipsis but rebasing reverted me to develop so I can split the css into its own pr as I reapply code.

PR #207, opened 2020-09-29
Replace jquery ready eventhandler

bitkeeper:
The current `www/footer.php` use at the bottom a two stage rocket:

- First; synchronous jquery files are loaded (very unexpected place to find the load of scripts by the way) .
- Second; an event handler on on document ready is registered with a `jQuery(document).ready` to create a pnotifier (this code is generated by a php method `ui_notify`)

When using gulp for bundling the jquery source all resources are marked with `defer` for loading, which breaks the functionality above. 
I replaced the code `jQuery(document).ready` by a generic multi plaform function `docReady`. Because the php function `ui_notify` make use of it I had to change it also.

### DIFF
--- a/www/footer.php
+++ b/www/footer.php
@@ -177,6 +177,86 @@
 <script src="js/jquery-ui-1.10.0.custom.min.js"></script>
 <!--endRemoveIf(USEBUNDLE)-->
 
+<script>
+	// Because with bundles jquery is loaded defered, the jQuery(document).ready(function() isn't availabe for the ui_notify construction.
+	// Solution is replace is by an other portable document ready event handler.
+
+	//Copied from https://stackoverflow.com/questions/9899372/pure-javascript-equivalent-of-jquerys-ready-how-to-call-a-function-when-t
+
+	(function(funcName, baseObj) {
+    // The public function name defaults to window.docReady
+    // but you can pass in your own object and own function name and those will be used
+    // if you want to put them in a different namespace
+    funcName = funcName || "docReady";
+    baseObj = baseObj || window;
+    var readyList = [];
+    var readyFired = false;
+    var readyEventHandlersInstalled = false;
+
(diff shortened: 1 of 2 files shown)


moodeaudio:
Is the routine the same as in the Git repo referenced in the S/O article (link below)?
https://github.com/jfriend00/docReady

If it is then the Git link is what should be referenced in your code comments because the code at that link contains an MIT license. I'll also need to download the sources from that repo (already did) and include it in our source tree in the /other directory.

bitkeeper:
You completely right. I didn't scroll down enough at StackOverflow to notice the code repo, my bad.
Importing the source and making the reference for such function is overkill fot this functionality, was just to lazy to code it out for the different browsers. 
I will find another solution.

moodeaudio:
This code is OK and we should use it.

All that needs changing in the PR is the comment line below. You can just delete the line because the code is same as in that Git repo and I'll be including the Git cosource in the /other directory of the source tree.

	//Copied from https://stackoverflow.com/questions/9899372/pure-javascript-equivalent-of-jquerys-ready-how-to-call-a-function-when-t

bitkeeper:
I think I can do better than current PR solution by :

- Instead of registering on document ready handler, just make it function.
- Call the function from application.js and script-panels.js. Central control of what the app does and when.
- Move the include of the jquery files at the bottom to the header.php. Central place where it is clear which resource are used.

After that it makes a little bit more sense and less tricks are needed.

moodeaudio:
That sounds good.

bitkeeper:
I did put the improved solution in a new pr #208.
Now you can chose what you like more.

Only one php statement left before it can be just a html template ... but enough refactoring for a little while

moodeaudio:
The alternate PR has been merged so feel free to close at your convenience.

PR #208, opened 2020-09-29
Replace jquery ready eventhandler (Variant 2)

bitkeeper:
The current www/footer.php use at the bottom a two stage rocket:

- First; synchronous jquery files are loaded (very unexpected place to find the load of scripts by the way) .
- Second; an event handler on on document ready is registered with a jQuery(document).ready to create a pnotifier (this code is generated by a php method ui_notify)
When using gulp for bundling the jquery source all resources are marked with defer for loading, which breaks the functionality above.

New implementation
- Instead of registering on document ready handler, just make it function.
- Call the function from application.js and script-panels.js. Central control of what the app does and when.
- Move the include of the jquery files at the bottom to the header.php. Central place where it is clear which resource are used.

After that it makes a little bit more sense and less tricks are needed.

### DIFF
--- a/www/footer.php
+++ b/www/footer.php
@@ -171,20 +171,7 @@
 	<span class="reconnect-msg">System shut down</span>
 </div>
 
-<!-- DISPLAY MESSAGES -->
-<!--removeIf(USEBUNDLE)-->
-<script src="js/jquery-1.8.2.min.js"></script>
-<script src="js/jquery-ui-1.10.0.custom.min.js"></script>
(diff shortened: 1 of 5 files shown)


moodeaudio:
I'm not seeing any notify popup's when making changes on the Config screens.

moodeaudio:
Nvm. Just needed a bit of code from scripts-panels pasted into scripts-configs.js. I'll push a commit.

PR #211, opened 2020-09-30
Switch from .html() to innerHTML

monkeypox:
for the big lists in scripts-library, it benchmarks significantly faster.


moodeaudio:
It's definitely a perf improvement :-)

Nice!

PR #215, opened 2020-10-01
Restore full old behavior of artist_album mode

scripple:
I don't have tagging to test it, but this should fully restore the old "album_artist" behavior.

### DIFF
--- a/www/js/scripts-library.js
+++ b/www/js/scripts-library.js
@@ -412,11 +412,17 @@ function filterAlbums() {
 	}
 	// Filter by artist
 	if (LIB.filters.artists.length) {
-        // @scripple:
-        var artistSongs = allSongs.filter(filterByArtist);
- 		var songKeys = artistSongs.map(function(a) {return a.key;});
- 		filteredAlbums = filteredAlbums.filter(function(item){return songKeys.includes(keyAlbum(item));});
- 		filteredAlbumCovers = filteredAlbumCovers.filter(function(item){return songKeys.includes(keyAlbum(item));});
+		// @scripple:
+		if (SESSION.json['library_tagview_artist'] == 'Album Artist') {
+			filteredAlbums = filteredAlbums.filter(filterByArtist);
+			filteredAlbumCovers = filteredAlbumCovers.filter(filterByArtist);
+		}
+		else {
+			var artistSongs = allSongs.filter(filterByArtist);
+			var songKeys = artistSongs.map(function(a) {return a.key;});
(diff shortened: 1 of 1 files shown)


monkeypox:
It might be worth making strict adherence optional. I don’t think many people use albumartist to categorize like the example on the forum and the general user base might benefit from seeing albums that the artist had songs on. The easiest way would probably be to just add a “Album Artist (strict)” option to the SESSION var.

scripple:
Fine with me.  I just figured the point of the switch was people wanted it to behave like it used to.  We had an edge case where it didn't.

moodeaudio:
@monkeypox,  You mean add a conditional around the code and have two separate options?
- Album Artist
- Album Artist [strict]

monkeypox:
Exactly. I’m thinking showing albums the artist shows on but isn’t album artist might help expose the user to works by the artist they might not otherwise see with just album artist.

It might also be worth thinking about moving the choice of what to show to a popup on the column headers instead of in the appearance dialog (still saving the state of course), though it’d be a trade off with the convenience of clearing the column with just one click (though tapping the menu header clears as well).

monkeypox:
Also in this case I think it’d just be checking for “album artist (strict)” rather than adding another conditional.

scripple:
This mod would only need to check “album artist (strict)” but the other check for "album artist" elsewhere in the code would have to check for both “album artist (strict)”  and “album artist” or a substring match or whatever.

moodeaudio:
@scripple, ok that will work.

@monkeypox, I think @the_bertrum also suggested that having the artist option somewhere on Tag view instead of tucked away in Appearance would be more convenient, but there are other options like the Library filter, etc. I think it depends on whether the option is obviously one that will be frequently changed by most users.

moodeaudio:
One thought is to add a cog icon to left of Library, similar to whats in Radio view, and migrate the Library section of Appearance to the cog.

monkeypox:
Maybe replace the clear cache item with a preferences one and move clear cache to a dialog with the other library specific prefs? The downside of a cog is it’ll make hitting the library menu harder because you’re not just hitting the screen edge.

moodeaudio:
Ya for sure there's a major downside to cluttering the views with buttons and cogs. Leaving it as-is prolly best.
Clear cache could be moved tho. I may have added it cos I was doing it so often testing all the Library stuff. Users would not need it that much.

scripple:
The other check could just be this of course.

    if (SESSION.json['library_tagview_artist'] != 'Artist') {
      return acc;
    }

PR #216, opened 2020-10-02
Fix lazyload image loading

monkeypox:
This circumvents an issue where lazyload would sometimes be called before the list had populated by adding a function that uses javascript promises to let moode know when an element exists, i.e. $.ensure(this).then(that).

It also:

#1 checks if the view is active before calling lazylode (looking at you playlist & radio)

#2 checks to see if lazyload has already been called and if so skips the actual lazyload call.

#3 allows a call to force lazyload to run anyway (e.g. for typedown search)

#4 allows setting skip_invisible (e.g. for typedown search)

#5 switch radio & playlist to use innerHTML like the two library lists

#6 adds an id to the play queue list markup to facilitate #5

#6 fixes a bug in renderui with the default cover image (was missing a '" ' at the end of the src.

#7 limits MPD.json['date'] to 4 characters which should maybe be done in the php but whatever

#8 adjusts the mobile playbar title text a little: slighty larger, shifted to the left a little.

#9 remove placeholder graphic from lazyload library, sets skip_invisible back to original default value of false and comments out srcset stuff we don't use.

### DIFF
--- a/www/css/moode.css
+++ b/www/css/moode.css
@@ -251,7 +251,7 @@ li.modal-dropdown-text:focus {color:#eee;}
 	.playlist span {margin-left:calc(3.8rem + 1vmin) !important;}
 	.playlist {padding:0 0 9rem 0;min-height:33vh;}
 	/*.playlist li:before {width:1.75em!important;}*/
-	.playlist .pl-action, .playlist .db-action a {padding:.4em 1.25em 1em 0;}
+	.playlist .pl-action, .playlist .db-action a {padding:.4em 1.25em 1em 0;margin-left:1.25em!important;}
 	.ui-pnotify {top:40% !important;}
 	#lib-albumcover {left:0;}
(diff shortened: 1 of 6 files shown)


moodeaudio:
Nice! Simpler code and a pretty massive perf improvement :-)

I'll go ahead and remove all the commented out code.

We still might need some sort of "Loading..."  indicator or backdrop for the edge cases where the PHP process actually takes a while. One such case would be if we offer the enable_m4a_probe thing. You can simulate a long lib load by uncommenting the line below in moode.php.

```
		case 'loadlib':
			//sleep(8); // To simulate a long Library load
			echo loadLibrary($sock);
                 	break;

```

monkeypox:
Maybe something a setinterval set to show the popup after a few seconds and have moode clear the interval and dismiss the popup after setting the global indicating the load has completed? That way it won’t show for fast loads but will for longer ones.

PR #225, opened 2020-10-09
Multiple Artists support

Phaet:
Please review my modifications to support multiple track artists.
1. The assumption is that album_artist should be never empty. To this end, in case it is empty, it is set either to artist if artist is a single value or otherwise to 'Unknown AlbumArtist'. In consequence, in scripts-library.js all constructs of the form album_artist || artist can be simplified to just album_artist, for example in sorting by AlbumArtist. (Sorting by Artist makes no sense when multiple Artists exist per track).
2. playerlist.php also cares now for (multiple) performer tags and adds them to the artists array. (Artist is regarded as generic term). (Note that also conductor would be in this category, but is not supported by mpd.)
3. All artist related functions are modified to support arrays of artists.
4. In tag view, when albums are filtered by selected artists, expanding/collapsing the track list of an album by clicking the album cover is extended to support multiple artists too.

### DIFF
--- a/www/inc/playerlib.php
+++ b/www/inc/playerlib.php
@@ -472,13 +472,39 @@ function genFlatList($sock) {
 				$item = count($flat);
 				$flat[$item][$element] = $value;
 			}
-			// @Atair: Gather possible multiple Genre values as array
+			// @Atair: Gather possible multiple Genre, Artist, and Performer values as array
 			elseif ($element == 'Genre') {
 				if ($flat[$item]['Genre']) {
 					array_push($flat[$item]['Genre'], $value);
 				}
 				else {
 					$flat[$item]['Genre'] = array($value);
+				}
+			}
+			elseif ($element == 'Artist') {
+				if ($flat[$item]['Artist']) {
+					array_push($flat[$item]['Artist'], $value);
+				}
+				else {
+					$flat[$item]['Artist'] = array($value);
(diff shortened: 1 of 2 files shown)


Phaet:
BTW: It's me @Atair  :-)
- Walther

moodeaudio:
Hi Walther,

I'll check it out this weekend :-)

Phaet:
Hi Tim,
I noticed that I missed a couple of commits since I started to work on the two files, so there are a lot of differences. Maybe I can rebase my edits on more recent files during the weekend.

moodeaudio:
That would be great if you could resolve the conflicts.

I'm not planning on making any mods to those files this weekend.

Phaet:
Still found a glitch which I need to fix.

Phaet:
I rebased (git rebase) the branch now to develop/HEAD thereby resolving all conflicts reported by git rebase.

Phaet:
Rebased and tested. Works smoothly in my test environment.

moodeaudio:
Nice!

I should be able to get to this in a few days. I'm just taking some time to polish up the new filtering feature and fix some minor bugs.

-Tim

Phaet:
Great! Thanks!

24.10.2020 16:15:17 Tim Curtis <notifications@github.com>:

PR #226, opened 2020-10-10
Support for 'Any' filter

monkeypox:
WIP as the filter section in the appearance dialog is going away but this allows a user to filter matching with mpd 'search any'.

Also:

#1 adds a !f bang command to the library search so you can quickly apply a filter which is the main way this is intended to be used.

#2 make setLibMenuHeader() show the current filter after the regular text in parenthesis

#3 added a call to setLibMenuHeader() in scripts-library after it renders the library so filtering gets the right header

#4 makes context menus min-width 12.5rem for spacing niceness

#5 Fixes the library menu (aka viewswitch) to use the normal font, adjusts the spacing to match the m menu and pads the text-only items to match the icon ones.

#6 removed a font size bump to mobile portrait, the text on the screen is alreay larger than the rest of the text on the screen, maybe we need to adjust if the font size setting has been bumped?

#7 fixed an error in adaptive backgrounds if it couldn't pull a color


monkeypox:
Don't merge this yet

PR #227, opened 2020-10-12
A few minor !f fixes

### LINKED ISSUE #1: What the SSH user and password is ?

monkeypox:
#1 - use a conditional for setLibMenuHeader() in loadLibrary() so it only applies for album & tag views

#2 - smarter application of (filter)

#3 - move cleartimeout to try to clear up an issue where it was showing up when switching tabs

#4 add a cleartimeout to filterhelp() to double good clear it out


monkeypox:
conflict #1 in playerlib is on crack, they're the same afaict, use the develop code
conflict #2 in playerlib is an optimization to avoid a var declaration, use swizzle
conflict #1 in scripts-library is my musicbrainz bypass commented out, use develop

moodeaudio:
Ok, I'll check em out

PR #228, opened 2020-10-14
Add Clear Filter to Library menu

monkeypox:
This shows a 'Clear Filter' menu item when a filter is active, selecting it clears the filter and hides the menu item.

Also:

#1 adds fa-undo for use with the new menu item

#2 don't need a conditional for menu-header output now, duh

#3 clears search reset button in filterhelp()

#4 !f by itself no longer clears the filter, by happenstance it shows the last filter which is kind of cool so I left it


monkeypox:
Someday I will learn rebase good

PR #242, opened 2020-10-29
Support for multiple artists and multiple genre in playback panel

Phaet:
Logic and layout rewritten.
Added artist in the list of extra meta data in preferences.
Function formatExtraTagsString in playerlib-nomin.js is obsolete an can be removed.


Phaet:
Please wait with merging.
I need to pin a bug and also need to clean up the code.

PR #243, opened 2020-10-30
[eqp] use q instead of bandwidth

bitkeeper:


moodeaudio:
This is doing Q <--> BW conversion in the UI so no need to recompile CAPS correct?

moodeaudio:
Nvm. I see the calc is done in PHP.

PR #246, opened 2020-11-02
Atair 7.0.4

Phaet:
Bug fix: When various album_artists in an album, clicking on the album did not expand the track list for some selected  album_artists.


Phaet:
My proposed getAlbumArtist function to set the album_artist prop is the better solution.

PR #249, opened 2020-11-05
Minor HTML corrections

gbh-uk:
Escaped html < and > characters where applicable.

indextpl.html - consolidated 2 "class" attributes into 1 & removed spurious </span> tag.

snd-config.html - corrected 2 id's.


monkeypox:
The “<“ and “>” characters didn’t need to be escaped as they’re used in a string but the change doesn’t really hurt anything aside from readability.

PR #254, opened 2020-11-08
Fixes/eqp false change notification

bitkeeper:


bitkeeper:
Closed because changes are revert manually.

PR #255, opened 2020-11-08
Allow moode to use monitors with non-Television resolutions.

scripple:
hdmi_group=1 <= TV resolutions only
hdmi_group=2 <= Computer resolutions only
hdmi_group=0 <= Either ok.

Why is it set to 1?


moodeaudio:
I don't recall why it's set to 1. Group =  0 would be best.

scripple:
Agreed and that's what this PR does.  Changes group to 0.

PR #256, opened 2020-11-08
Combine track info and audio info into one dialog

monkeypox:
This adds a tab bar to the audio info dialog with one showing track info and the other the hardware related info.

Changes:

audioinfo.php

make "n/a" items (e.g. for crossfade if airplay is active) say "off"

simplify resample list to better fit on one line. Upsampling is implied, but I left the resample text as it's the only one that can go in either direction.

audioinfo.html

add tabs for track/station & hardware

add div for track info

only show active info by adding class of the php var to the li of items that might have "off" or "None" as their data so they can be hidden.

moode.css

remove borders from viewswitch-cfg tabs

add audioinfo-tabs rule

fix no bottom margin on currentartist for mobile

bump mobile modal-body l/r padding a bit

panels.css

commented out .playback-controls as it isn't used

add a max-width to img.coverart so we can support smaller landscape sizes easier but i think we should have a minimum supported screen size (raspbery pi touchscreen maybe?)

modify/add audio info dialog rules

scripts-library.js

switch to new style track info call

scripts-panels.js

switch to new style track info call

add click handlers for the audio info tab bar

playerlib.js

switch to new style track info call

bump modal backdrop opacity a bit

add audiohardware function to handle the menu item from header.php

add audioinfo function that handles the display of the audio info dialog

generalize the itemInfoModal function so it can work with different elements, this is only called by audioinfo right now but might be useful elsewhere down the road. Possible enhancement would be to detect an empty list (e.g. if no track is playing or paused) and report something like "no information" as the output but the other half of the audio info dialog has empty strings so ¯\_(ツ)_/¯.

### DIFF
--- a/www/audioinfo.php
+++ b/www/audioinfo.php
@@ -214,20 +214,20 @@ else {
 // Renderers
 if ($_SESSION['airplayactv'] == '1' || $_SESSION['spotactive'] == '1' || $_SESSION['slactive'] == '1' || $_SESSION['inpactive'] == '1' || $btactive === true) {
 	$resample_rate = '';
-	$resample_quality = 'n/a';
-	$polarity_inv = 'n/a';
-	$crossfade = 'n/a';
-	$crossfeed = 'n/a';
-	$replaygain = 'n/a';
-	$vol_normalize = 'n/a';
+	$resample_quality = 'off';
+	$polarity_inv = 'off';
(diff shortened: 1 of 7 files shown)


monkeypox:
please resolve manually as github has special needs 🥴

Untitled 2

PR #259, opened 2020-11-19
Add some FluxFM radio stations.

e11bits:
source https://www.fluxfm.de/radio-livestream/


moodeaudio:
The new radio station feature in moOde 7 requires metadata for each station that is part of the built-in station package. The metadata is stored in the cfg_radio table. https://github.com/moode-player/moode/blob/develop/var/local/www/db/moode-sqlite3.db.sql

e11bits:
I've added the entries to the DB script. The genres might not fit 100%

PR #260, opened 2020-11-19
Fix Bluetooth agent for devices requiring PIN

gabibbo97:
This hardcoding of the pin to the 0000 screen solves pairing issues for some Xiaomi devices that require a PIN code to associate

### DIFF
--- a/www/command/bt-agent.py
+++ b/www/command/bt-agent.py
@@ -78,11 +78,11 @@ class Agent(dbus.service.Object):
 
     @dbus.service.method(AGENT_INTERFACE, in_signature="o", out_signature="s")
     def RequestPinCode(self, device):
-        raise Cancelled("Pin code not supported")
+        return "0000"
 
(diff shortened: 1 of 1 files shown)


moodeaudio:
How does this change affect other devices?

gabibbo97:
So in my testing with two Android devices (Redmi Note 9 Pro / Redmi 8A) I only managed to pair the devices with the patch that hardcodes the pin to 0000, before it was failing with something along the lines of "pairing refused".
On my Fedora 33 machine, with the patch I can connect and cast audio
image

moodeaudio:
The challenge for me is that I don't really have the expertise in this part of bluetooth to know whether this change will be ok not.

Maybe someone can confirm that IOS devices work ok with this change?

gabibbo97:
Me too, let's wait until someone with different devices chimes in

moodeaudio:
Closed due lack of response

PR #261, opened 2020-11-30
Atair 7.0.5

Phaet:
Added preference to allow for selection of the search site for artist/album search.
New search sites in addition to Google are Bing, DuckDuckGo, Ecosia, Startpage, Yahoo, Musicbrainz and Discogs.

The new param to add is:
moodeutl -q "INSERT INTO cfg_system (id, param, value) VALUES (152, 'search_site', 'Google')"


moodeaudio:
What are the advantages over just a simple Google search?

Phaet:
Maybe this was due to my personal paranoia: I prefer more privacy search engines like DuckDuckGo and avoid Google as far as possible.
Maybe others have the same feelings about Google and want to decide themselves which search engine to use.
You can argue whether this maybs really a benefit (to me is), but providing the more specific search in Musicbrainz and Discog clearly is.

Phaet:
Initially I also had Wikipedia in the list but the search result is only useful for artist, but in combination with the album title. So I decided to remove it.
But it is still mentioned in the quick help ...

moodeaudio:
Hi,

Please fix conflict and then I'll merge and test.

-Tim

PR #262, opened 2020-12-03
Introduce preference setting for search site (Second take :-) )

Phaet:
Introduce preference to allow the user to select the search site out of Google, Bing, DuckDuckGo, Ecosia, Yahoo, Musicbrainz, Dicogs and Wikipedia.

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -959,7 +959,38 @@ function genSearchUrl (artist, title, album) {
     		searchStr = artist + '+' + album;
     	}
 
-        var searchEngine = 'http://www.google.com/search?q=';
+        var searchEngine = '';
+	switch (SESSION.json['search_site']) {
+		case 'Google':
+			searchEngine = 'http://www.google.com/search?q=';
+			break;
+		case 'Bing':
+			searchEngine = 'http://www.bing.com/search?q=';
+			break;
+		case 'DuckDuckGo':
+			searchEngine = 'http://www.duckduckgo.com/?q=';
+			break;
+		case 'Yahoo':
+			searchEngine = 'http://search.yahoo.com/search?p=';
+			break;
+		case 'Ecosia':
+			searchEngine = 'http://www.ecosia.org/search?q=';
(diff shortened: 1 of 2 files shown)


gbh-uk:
If I may make a suggestion - to add an option 'None' - that when selected disables the link to the search engines.
External search is a feature that I never use, and it's a pain when using the official Pi touch screen and someone clicks the link then asks 'how do I get back'.

Phaet:
I named the option "Disabled" just to emphasize the effect.

moodeaudio:
Nice feature add :-)

it might be better to list it under the Playback section of prefs since it's only relevant in Playback view.

Phaet:
That's right. Do you move the item?

moodeaudio:
Yes, I'll make the changes. There are a few other minor tweaks for example the search site list should prolly have "Disabled" first and then the rest of the items in alphabetic order. One other modification that needs to be made is in the MusicBrainz and Wikipedia cases where the artist and album vars are used. These vars will be undefined when radio station is playing. Its just a matter of parsing out those values from the searchStr.

Phaet:
Oups, this case I missed in my tests.
I would simply add two lines in to define album/artist in case:
if (typeof(artist) === 'undefined' || artist === 'Radio station') {
		var searchStr = title.replace(/-/g, ' ');
		searchStr = searchStr.replace(/&/g, ' ');
		searchStr = searchStr.replace(/\s+/g, '+');
     artist = searchStr;
     album = '';

moodeaudio:
For radio we only have "artist - title" to work with so for the MB and Wiki searches we can only use artist.

Phaet:
I was just about to send a new PR with the corrections/mods, but noticed that you overtook me on the fast lane ... :-)

moodeaudio:
Reminds me of "Life in the Fast Lane" by the Eagles.

Phaet:
Yeah, from Hotel California. Must have been in the late 70th. Reminds me how old I am ... 😁

PR #263, opened 2020-12-04
Fix missing brackets at if condition

bitkeeper:
While performing a gulp build found missing brackets arround javascript condition.


moodeaudio:
Already fixed as part of an earlier commit this morning.
https://github.com/moode-player/moode/commit/c73bb4bc40a3d56a99cf28c42101c5444d2a011f

Go ahead and close at your convenience :-)

PR #266, opened 2020-12-06
Fixes/eqp improvements

bitkeeper:
- [fixed] getActivePresetIndex() exception when not preset is active
- [feature] rename the enter button to new and moved it next to the SAVE and REMOVE button
image
- [feature] when a new curve is create base is on the current selected (was current active)
- [fixed] save after new curve was created showed/selected the active curve

### DIFF
--- a/www/eqp-config.php
+++ b/www/eqp-config.php
@@ -28,6 +28,7 @@ $dbh = cfgdb_connect();
 $eqp12 = Eqp12($dbh);
 
 $curve_config = NULL;
+$_selected_curve_id = NULL;
 
 function postData2Config($data, $bands) {
 	$config = [];
@@ -49,6 +50,7 @@ if (isset($_POST['save']) && $_POST['save'] == '1') {
 	$curve_id = intval($_POST['curve_id']);
 	$config = postData2Config($_POST, 12);
 	$eqp12->setpreset($curve_id, NULL, $config);
+	$_selected_curve_id = $curve_id;
 
 	if($curve_id == $eqp12->getActivePresetIndex() ) {
 		$playing = sysCmd('mpc status | grep "\[playing\]"');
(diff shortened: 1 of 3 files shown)


moodeaudio:
There is still some breakage.

- New curve does not appear in the list after adding it
- The message "New curve added" does not appear

There is also some oddness in the eqfa12p worker job. Its missing an mpd restart which is needed after changing the curve or turning it off. I'll go ahead and add that.

bitkeeper:
Yes is see that when the alsaequal and eqp case use the same imp for the worker the restart was gone ... oeps.

moodeaudio:
I can remember how it was originally but currently the Graphic EQ does not need an mpd restart after changing curve, its only the PEQ that needs the restart. Anyway you will see the commit for worker.

bitkeeper:
Ok the new curve is only create when you select a curve first (so curve=xx is present as GET parameter).
Will fixes this.

moodeaudio:
There are still some issues.

- NEW only works the first time. After that it does not add a new curve unless you return to Audio Config and then back to EQP Config. We should be able to continue to add or remove curves in EQP Config without having to return to Audio Config each time.

bitkeeper:
With PR #268 I don't have see any problems: After first enter config I can press the new button serveral times.
Can you please provide the steps to reproduce ?

bitkeeper:
What I can reproduce is something different:
EQP active with non default curve
* Go to the EQP config 
* (it shows the active, non default profile )
* Press new
* New curve is created
Only it is based on the default, which should be the active one 

If you select it from the dropdown list in the config page it will be based on the correct one.

moodeaudio:
To repro the issues

1. Audio Config, leave EQ Off
2. EDIT EQ curve
3. NEW curve "Bob"
4. NEW curve "Frank"

At this point "Frank" is displayed in the field

5. REMOVE "Frank"

The message "Curve removed" does not appear and "Frank" is still in the list.
"Default curve" is displayed in the field

6. Select "Frank" from the list
7. REMOVE "Frank"

This succeeds

8. NEW curve "Betty"

The message "Curve added" does not appear and "Betty" is not added to the list.

bitkeeper:
Ok clear, this is some behaviour inherited from the original eqp4 implementation.
Has to do with that in that case the curve is missing as GET parameter. 
Nicest would be redirect after an new or deleted curve to get the curve parameter back.

Probally will chose to short way by adding hiden field to the form(s). Will fix it tomorrow evening, unless you're ahead of me ;-)

moodeaudio:
No rush. I have plenty to do :-)

moodeaudio:
Graphic EQ works a bit differently in that newly added curve must be saved first to become persistent. Then no problems deleting etc and no need for hidden fields.

I think we should try to get PEQ to work same way.

PR #270, opened 2020-12-08
Add camilladsp and alsa_cdsp to moode build recipe

bitkeeper:
Not sure where to leave in the file, now placed it just after the Alsaequal/eqp part?

Added as well how to build from source as how to install the prebuilds in the repo.


moodeaudio:
That works.

PR #279, opened 2020-12-24
CamillaDSP integration

JWahle:
This is very simple CamillaDSP integration:
- adds an On/Off switch in Configure > Audio > Equalizers > CamillaDSP
- below the switch is an indicator, if the config exists, is valid or invalid (with error message)
- the info button contains a sample config that would use the currently selected output device and instructions on what to do and where to look, if things go wrong. (Currently the config has to be uploaded to /home/pi/camilladsp.yml so it can be easily uploaded via scp)


moodeaudio:
Thanks! :-) will check it out after the holidays.

moodeaudio:
I added your ID to the Test Team sub-forum. Refresh your Browser and it should appear.

-Tim

JWahle:
Cool - that was quick. Thanks.

Issue #280, opened 2020-12-24
Genius lyrics API


Stephanowicz:
Hi,

it would be a nice option to have a lyrics query for the song actually playing.
The [Genius lyrics API](https://docs.genius.com/#/getting-started-h1) is free to use and provides many songlyrics.

I have searched for an easy implementation for querying and finally created a php script. Nevertheless one still needs to install the xml-php lib `sudo apt-get install php7.3-xml`, as xpath is used for parsing the lyrics-page. For the initial request I used a function by [Allen Fair](https://gist.github.com/afair/a7c7adc52b7b49bf362935e665a87633) - I didn't change it here, but there's lots of unnecessary stuff that can be removed.

In order to use the API one needs to create a free account at  [genius.com](https://genius.com/signup_or_login) - then You need to [create a new api-client](https://genius.com/api-clients) in order to generate a client access token.
This token has to be put in the first line:

```
<?php
$ClientAccessToken = '<<<PUT YOUR CLIENT ACCESS TOKEN HERE>>';
$ARTIST_=shell_exec('mpc --format %artist% | head -n 1');
$ARTIST_=str_replace("\n","",$ARTIST_);
$ARTIST=str_replace(" ","+",$ARTIST_);

$TITLE_=shell_exec('mpc --format %title% | head -n 1');
$TITLE_=str_replace("\n","",$TITLE_);
$TITLE=str_replace(" ","+",$TITLE_);

function http_request($url, $opt=array()) {
/**
 * Makes a remote HTTP request and returns the response.
 *
 * @author    Allen Fair <allen.fair@gmail.com>
 * @copyright 2018 Allen Fair
 * @license   https://opensource.org/licenses/MIT  MIT
 * @code      https://gist.github.com/afair/a7c7adc52b7b49bf362935e665a87633
 *
 * @param string  $url     The full URL endpoint: "https://example.com/endpoint"
 * @param array   $opt     An Associative Array of extended request options
 *
 *        string  'method'       one of: GET, POST, PUT, PATCH, DELETE, HEAD
 *        string  'url'          URL override
 *        string  'endpoint'     Optional endpoint appended to URL
 *        string  'user'         user name to send for HTTP Basic Authorization
 *        string  'token'        creates an "Authorization: Bearer" header
 *        string  'accept'       creates an Accept: header for the mime/type
 *        string  'password'     password to send for HTTP Basic Authorization
 *        array   'headers'      array of "Header-Name: value", ...
[...]

PR #282, opened 2020-12-29
Display song-lyrics for current song

Stephanowicz:
Added a lyrics-query for the current playing song.  (for more details, please take a look [here](https://github.com/moode-player/moode/issues/280))
(Radio stations haven't been taken in consideration - maybe this needs additional steps.)

Requirements:

- Genius client access token
- php-dom (`sudo apt-get install php7.3-xml`)

The query is using the [genius.com api](https://docs.genius.com/#/getting-started-h1) - to use it You'll need to create a [free account](https://genius.com/signup_or_login). Then You need to [create a new api-client](https://genius.com/api-clients) in order to **generate a client access token**.
At moment, this token has to be put in the first line of the ./command/geniuslyrics.php file. 

The modal lyrics-window can be accessed by the mainmenu and in the playbar-menu.

Todo:

- An option to enter the token on the settings page (maybe with link for account creation and creation of api-clients).
- An option to change the search parameters on the lyrics window - in case there was no or wrong result
_- Open the lyrics-window immediately an show a wait spinner till the text displays as it may take a few seconds to get a result._

Cheers, Stephan


www/footer.php
```diff
@@ -138,6 +138,20 @@
 	</div>
 </div>
 
+<!-- Lyrics -->
```
bitkeeper, line 141: Maybe a little bit overkill to add it also to the main menu ? I think on the play context menu would be enough.

www/js/scripts-library.js
```diff
@@ -1405,6 +1405,12 @@ $('#context-menu-playback a').click(function(e) {
             $('#setfav-modal').modal();
         });
 	}
+  else if ($(this).data('cmd') == 'song-lyrics') {
```
bitkeeper, line 1408: This handler cause a double load on my system, because the one from playerlib.js also triggers ?

www/js/playerlib.js
```diff
@@ -2271,7 +2271,13 @@ $('.context-menu a').click(function(e) {
 		$('#quickhelp').load('quickhelp.html');
         $('#quickhelp-modal').modal();
     }
-
+    //Song-lyrics
+    else if ($(this).data('cmd') == 'song-lyrics') {
```
bitkeeper, line 2275: What about replacing this by:

```
		$('#lyrics').load('lyrics.php?content=loadingmessage');
		$('#lyrics-modal').modal();
		$('#lyrics').load('lyrics.php?content=lyrics');
```

www/lyrics.php
```diff
@@ -0,0 +1,39 @@
...
+$cmd = 'php /var/www/command/geniuslyrics.php';
+
+passthru('sudo ' . $cmd, $text);
+
+//$fh = fopen($file, 'r');
+//$text = fread($fh, filesize($file));
+//fclose($fh);
+
+//$tpl = 'lyrics.html';
+//eval('echoTemplate("' . getTemplate("templates/$tpl") . '");');
```
bitkeeper, line 38: Remove the comment and maybe replace the current construction by:
```
if( isset($_GET['content']) && $_GET['content'] == 'lyrics') {
    $cmd = 'php /var/www/command/geniuslyrics.php';
    passthru('sudo ' . $cmd, $text);
    echo $text;
}
else {
    $tpl = 'lyrics.html';
    eval('echoTemplate("' . getTemplate("templates/$tpl") . '");');
}
```
This in combination with the suggestion for playerlib.js

bitkeeper:
@moodeaudio
Is it ok if the song lyrics action is always present in the play context menu or should it not be visible when no token is set ?
The config option to set the token; on which config page should it be ?

Stephanowicz:
@bitkeeper 

Thank You!

ok - let's see:
`On the end of the lyrics I always get the number 0 ?` 
yes - I will remove that - I think it's the return code from 'passthru'

`You used two integration point in the UI; the main menu and the play context, I think the song context would be enough.
`  
I think it's necessary as  (at least on my dektop) the main menu is not available when the play-bar is visible in the cover-view mode and vice versa - (both are only visible same time on the library-view)

`When loading the lyrics from the play context the lyrics are loaded twice.`
Hmm - sorry I cannot reproduce that, and it shouldn't happen as these are eventtriggers of the menuactions of the menu where they are declared...  but I made some changes I didn't push yet - maybe You try again after I uploaded them

`If no valid key is provided the message is shown that no lyrics are found. Is it possible to detect that the token is incorrect and show that ?`
Yes, this needs a better handling of the response from the genius server.

`Maybe you can do the double loading with the template.html and lyrics.php with a single page with a GET argument, that looks a little bit cleaner.`
Well, these are remainders of different approaches - will clean it up 

I was a bit lazy updating my code changes - will do that soon and remove the unnecessary stuff.

`Is it ok if the song lyrics action is always present in the play context menu or should it not be visible when no token is set ?
The config option to set the token; on which config page should it be ?`
been thinking about that, too - that would be favourable

bitkeeper:
> think it's necessary as (at least on my dektop) the main menu is not available when the play-bar is visible in the cover-view mode and vice versa - (both are only visible same time on the library-view)

image

Not sure if I understand you well ? The play context menu is available everywhere; library, tag view, play and cover ? So why you need the main menu entry ?

moodeaudio:
Correct, the context menu button ... is available in all the play views so no need to have anything on the Main menu.

Stephanowicz:
ok - thank You.

seems like the bottom gets cut of on my desktop xD

image

seems like I have to start a ticket ;)

www/header.php
```diff
@@ -178,7 +178,6 @@
 					<li><a href="blu-config.php"><i class="fas fa-wifi sx"></i> BlueZ</a></li>
 					<li><a href="javascript:$('#players-modal .modal-body').load('players.php',function(e){$('#players-modal').modal('show');}); void 0"><i class="fas fa-forward sx"></i> Players</a></li>
                     <li><a href="javascript:audiohardware()"><i class="fas fa-music sx"></i> Audio info</a></li>
-          <li class="context-menu"><a href="#notarget" data-cmd="song-lyrics"><i class="fas fa-bars sx"></i> Song lyrics</a></li>
```
Stephanowicz: remove lyrics item from main menu as it is always accessible from context-menu

www/js/playerlib.js
```diff
@@ -2271,13 +2271,6 @@ $('.context-menu a').click(function(e) {
 		$('#quickhelp').load('quickhelp.html');
         $('#quickhelp-modal').modal();
     }
-    //Song-lyrics
-    else if ($(this).data('cmd') == 'song-lyrics') {
-      $('#lyrics').load('./templates/lyrics.html');
-      $('#lyrics-modal').modal();
-      $('#lyrics').load('lyrics.php');
-  //  $('#lyrics').load('lyrics.php',function(e){$('#lyrics-modal').modal('show');});
-    }
```
Stephanowicz: remove lyrics item from main menu as it is always accessible from context-menu

www/lyrics.php
```diff
@@ -22,18 +22,6 @@
...
-
-//$fh = fopen($file, 'r');
-//$text = fread($fh, filesize($file));
-//fclose($fh);
-
-//$tpl = 'lyrics.html';
-//eval('echoTemplate("' . getTemplate("templates/$tpl") . '");');
-echo $text;
+$cmd = 'php ' . dirname(__FILE__) . '/command/geniuslyrics.php';
+$result = shell_exec($cmd);
```
Stephanowicz, line 26: cleaned up the code
changed system call from `passthru` to `shell_exec` - so we have no return code from the call (eg '0')

www/js/scripts-library.js
```diff
@@ -1584,3 +1581,9 @@ function formatLibTotalTime(seconds) {
 function formatNumCommas(x) {
     return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
 }
+
```
Stephanowicz, line 1584: created an own function for the menu click event of the lyrics item.

www/command/geniuslyrics.php
```diff
@@ -191,16 +191,26 @@ function http_request($url, $opt=array()) {
...
-//
+  else{
+    if($result[0]=="401"){
+      echo "401 Unauthorized<br /><br />please check Your client accesstoken!<br /><br />";
+    }
+    else{
+    	echo "<br /><br />Sorry, no lyrics found...<br /><br />";
+    };
+  }
 }
```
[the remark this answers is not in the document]
moodeaudio: I see the included http_request() function is MIT licensed but there is no license attached to the rest of the code in the file.

Stephanowicz:
Ok, nginx now seems to accept the direct loading of geniuslyrics.php after surpressing warnings and notices in it.
`error_reporting(E_ALL & ~E_NOTICE & ~E_WARNING);`

so the lyrics.php is kinda obsolete and one can do:
`$('#lyrics').load('./command/geniuslyrics.php');`

This is still in progress as I'm thinking of adding more options like refining the query if no result is returned, by entering a different query-string - this would also be useful if there is not enough info from the audio file - like missing title or missing artist.
This would also involve  to pass arguements like title and artist to the function.
The handling of the client access token would also happen there.

Btw: Is it ok to store the token in the database? Or is there another option to save this?
And if the database is the recommended place - in which area?
also want to revive Marcel's question: what would be a good place in the settings area to enter the token?

Cheers, Stephan

Stephanowicz:
as announced I removed the lyrics.php - geniuslyrics.php is loaded directly in js.
Also added params to the function - so it can be called with different params.
e.g. if a song has no artist You now may enter the artist manually. 

image

moodeaudio:
I'll have to read the TOS for genius and see if there are restrictions on use of access tokens. Most of these "free" services have a bunch of legal restrictions.

Does all your code have a license attached?

bitkeeper:
> as announced I removed the lyrics.php - geniuslyrics.php is loaded directly in js.

I did like your previous implementation more with the separate `geniuslyrics.php` and `lyrics.php`:

1. The lyrics.php should be the adapter between the lyrics provider and the moOde frontend.
2. And `geniuslyrics.php` can just do in what it is good; providing the lyrics. 3. 

I clearly separate concerns. And in case the other lyrics provider are required or genius is oboslute, there is one central place to handle it. Also things like extracting the token from the db or anyother info from moOde should be a lyrics.php issue , just to pass it as argument to geniuslyrics.php.

I am strongly in favour or reverting this change.

moodeaudio:
Prolly a good idea in case we ever need to use a different lyrics service.

Stephanowicz:
your wish is my command... ;) 

well, I think it's a bit more complicated but it's working.

Cheers, Stephan

Stephanowicz:
> I'll have to read the TOS for genius and see if there are restrictions on use of access tokens. Most of these "free" services have a bunch of legal restrictions.
> 
> Does all your code have a license attached?

Well, this part makes it difficult:

`Except as expressly authorized by Genius in writing, you agree not to modify, copy, frame, scrape, rent, lease, loan, sell, distribute or create derivative works based on the Service or the Genius Content, in whole or in part, except that the foregoing does not apply to your own User Content (as defined above) that you legally upload to the Service. In connection with your use of the Service you shall not engage in or use any data mining, robots, scraping or similar data gathering or extraction methods. Any use of the Service or the Genius Content other than as specifically authorized herein is strictly prohibited. As between you and Genius, the technology and software underlying the Service or distributed in connection therewith is the exclusive property of Genius, our affiliates and our partners (the "Software"). You agree not to copy, modify, create a derivative work of, reverse engineer, reverse assemble or otherwise attempt to discover any source code, sell, assign, sublicense, or otherwise transfer any right in the Software. Any rights not expressly granted herein are reserved by Genius.`

Stephanowicz:
Sorry - the last commit shouldn't show up here - I wanted to start a new pull request...
can this be undone or how can I start a second pull request?

Nevertheless - I added the "comment" tag to the audiofile info - I thought I read somewhere that it has been already added...?
It's just a really little addition - I dunno wich files provide the comment tag - I know of flac and ogg 

Cheers,

Stephan

moodeaudio:
The TOS seems to prohibit using code to extract the lyrics. Very odd since how else are the the lyrics data going to read by people other than by some software code.

But even if the TOS allows your code modules to be used some of the code does not have a license attached and so can't be considered for merging anyway.

moodeaudio:
There are still issues with this PR that have to be resolved before I can can be considered.

Several new files have no license header or copyright attribution

lyrics.html - missing license header for the file
lyrics.php - no copyright attribution designating the author of this code
playerlib.php - should not be in this PR since there does not appear to be any code thats used. Also it looks like an older version of the file.
MIT.license - this file should not be in the PR since it has no purpose
geniuslyrics.php - this appears to use an MIT licensed function as-is but also contains additional code outside the function that has no attached license.

Stephanowicz:
Ok, I hope I did it right now :D

I'm not so used to this license stuff...

moodeaudio:
Ok thanks, all files have licenses.

Are the files below needed in the PR?

playerlib.php
playerlib.js

Stephanowicz:
> Ok thanks, all files have licenses.
> 
> Are the files below needed in the PR?
> 
> playerlib.php
> playerlib.js

Nope
 - in the .js I had the mainmenu changed but then restored it as only one entry is needed, which is in the player context menu
 - in the .php I made the addition for the "comment" tag - but I didn't manage to let it not show up here....

moodeaudio:
I can merge and test after they are removed.

Stephanowicz:
LOL ... really sorry for the mess!
I think I'll better create a new request

PR #284, opened 2020-12-30
Features/camilladsp improvements

bitkeeper:
Added to the CamillaDSP settings page:

- Protect remove of active configuration
- When active configuration is upload reapply patch device and format
- When other configuration is selected at the pipeline configuration check the config always (still manual check is also available(
- Added Info button to convolution file list; provide mediainfo about the convolution file (if it is a soundfile).


JWahle:
@moodeaudio I did not test this, but the changes look fine to me. I suggest you merge and we test later.

PR #286, opened 2020-12-31
Update watchdog.sh

atatevyan:
This is to fix the false error appearing in moode.log:

`watchdog: Error: PHP session permissions
`

Basically, we should wait until 'sessionid' is returned from DB to avoid malformed SESSION_FILE variable.


moodeaudio:
The watchdog error is caused by order of execution issue in worker.php which causes the sql query in watchdog to occasionally fail because the cfg_system table is being updated at that time. The fix is to do the update before launching watchdog.

I've fixed it worker.php but if you still experience failures I'll reopen the PR.

atatevyan:
Nevertheless, it would be a good idea to add a file existence check to watchdog.sh before testing its permissions. Otherwise, if the sessionid is by chance not returned, then the permission test block will erroneously evaluate to True.

PR #296, opened 2021-01-08
Comment tag for audioinfo

Stephanowicz:
Added the "comment" tag for the audio info page.
Some audiofiles support the comment tag - like ogg and flac.

This might be usefull to add additional infos like the setup of the band, the recording location and so on.

image

Cheers, Stephan


moodeaudio:
This is intermingled with another PR and so not possible to merge.

I think its already on the the TODO list for 7.1.0 though so I can make the change manually on my end. There is also a need to check the Library preferences setting "Include comment tag" in order to be able to display it in Track info.

PR #297, opened 2021-01-08
[CamillaDSP] Add quick way to use a convolution filter

bitkeeper:
Provide a quick way to set a simple filter pipeline, if as configuraiton 'Quick convolution filter' is choosen.

image

The available settings:

- gain
- impulse response to use for left
- impulse response to use for right
- the source type of the impulse response

The used pipeline looks like:
image


bitkeeper:
@moodeaudio I think this one is ready.

PR #298, opened 2021-01-09
Added option to set max ALSA volume in dB

### LINKED ISSUE #275: HDMI output clipping with volume control disabled
I just tested moode via HDMI and got some subtle but sometimes very audible distortion.
My volume is set to "Disabled (0dB)".

This seems to be due to the ALSA volume being at 100% (+4dB).
If I set it to 0dB via `amixer set HDMI 0`
I get clean output.
In `alsamixer` I can only set the volume percentage to 86% (-0.01dB), so setting the volume this way is not optimal, if bit perfect output is desired.

However, at the next start, the volume is back to 100%.
Could you make sure that HDMI volume is set to 0dB and stays there when volume control is disabled?

JWahle:
Fixes #275 

I added a field to set the volume in dB.
If one is set, the other is updated accordingly.
If setting dB fails for some reason, the % value stays and dB value becomes blank.

set_volume_in_dB

I also thought about adding some indicator, which clarifies whether % or dB value is used to set the volume.
Since setting the % usually gives a fractional dB value, it should be obvious, which one is currently in use.
However, if you know a simple, pretty solution to this, I will implement it.
I tried with simple checkboxes, but didn't like the look of it:
set_volume_in_dB_with_checkbox

What do you think?

### DIFF
--- a/var/local/www/commandw/slpower.sh
+++ b/var/local/www/commandw/slpower.sh
@@ -21,15 +21,13 @@
 
 SQLDB=/var/local/www/db/moode-sqlite3.db
 
-RESULT=$(sqlite3 $SQLDB "select value from cfg_system where param='alsavolume_max' or param='alsavolume' or param='amixname' or param='rsmaftersl' or param='wrkready' or param='inpactive'")
+RESULT=$(sqlite3 $SQLDB "select value from cfg_system where param in ('alsavolume', 'rsmaftersl', 'wrkready', 'inpactive')")
 # friendly names
 readarray -t arr <<<"$RESULT"
-ALSAVOLUME_MAX=${arr[0]}
-ALSAVOLUME=${arr[1]}
-AMIXNAME=${arr[2]}
-RSMAFTERSL=${arr[3]}
-WRKREADY=${arr[4]}
-INPACTIVE=${arr[5]}
+ALSAVOLUME=${arr[0]}
+RSMAFTERSL=${arr[1]}
+WRKREADY=${arr[2]}
+INPACTIVE=${arr[3]}
 
 if [[ $INPACTIVE == '1' ]]; then
 	exit 1
@@ -59,7 +57,7 @@ if [[ $WRKREADY == "1" ]]; then
 		sleep 1
 		$(sqlite3 $SQLDB "update cfg_system set value='1' where param='slactive'")
 		if [[ $ALSAVOLUME != "none" && $VOPT == "" ]]; then
-			/var/www/command/util.sh set-alsavol "$AMIXNAME" $ALSAVOLUME_MAX
+			/var/www/command/util.sh set-alsavol-to-max
 		fi
 	else
 		# Value 2 is returned
--- a/var/local/www/commandw/spotevent.sh
+++ b/var/local/www/commandw/spotevent.sh
@@ -21,14 +21,12 @@
 
 SQLDB=/var/local/www/db/moode-sqlite3.db
 
-RESULT=$(sqlite3 $SQLDB "select value from cfg_system where param='alsavolume_max' or param='alsavolume' or param='amixname' or param='mpdmixer' or param='rsmafterspot' or param='inpactive'")
+RESULT=$(sqlite3 $SQLDB "select value from cfg_system where param in ('alsavolume', 'mpdmixer', 'rsmafterspot', 'inpactive')")
 readarray -t arr <<<"$RESULT"
-ALSAVOLUME_MAX=${arr[0]}
-ALSAVOLUME=${arr[1]}
-AMIXNAME=${arr[2]}
-MPDMIXER=${arr[3]}
-RSMAFTERSPOT=${arr[4]}
-INPACTIVE=${arr[5]}
+ALSAVOLUME=${arr[0]}
+MPDMIXER=${arr[1]}
+RSMAFTERSPOT=${arr[2]}
+INPACTIVE=${arr[3]}
 
 if [[ $INPACTIVE == '1' ]]; then
 	exit 1
@@ -43,7 +41,7 @@ if [[ $PLAYER_EVENT == "start" ]]; then
 	$(sqlite3 $SQLDB "update cfg_system set value='1' where param='spotactive'")
 
 	if [[ $ALSAVOLUME != "none" ]]; then
-		/var/www/command/util.sh set-alsavol "$AMIXNAME" $ALSAVOLUME_MAX
+		/var/www/command/util.sh set-alsavol-to-max
 	fi
 fi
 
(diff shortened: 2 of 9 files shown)


moodeaudio:
Whats the usage scenario for having two different volume scales for Max ALSA vol ?

The only dB value that works in amixer sset command is "0dB". It won't accept negative values and so it's sort of useless as a general method for setting volume. Also using a dB scale would require knowing what the the audio device's hardware volume scale was so the UI volume knob could be scaled correctly.

JWahle:
> Whats the usage scenario for having two different volume scales for Max ALSA vol ?

The usage scenario is the one from #275:
You cannot set exactly 0dB on the Pi's HDMI output using percentages (96% => -0.25dB)

> The only dB value that works in amixer sset command is "0dB"

It is possible to set negative values as [explained here](https://github.com/moode-player/moode/issues/275#issuecomment-747123899)

> Also using a dB scale would require knowing what the the audio device's hardware volume scale was so the UI volume knob could be scaled correctly.

I don't think this is relevant. This PR does not change the scale of the volume knob. The "Max ALSA volume" setting is (and was) only used, when MPD volume is set to software or disabled.

JWahle:
There is a different approach for the UI that I want to try out before merging.

moodeaudio:
Ok the -- means "no more options" would allow -dB values.

So this is to remedy the 100% = +4dB for the HDMI and Headphone ports?

The challenge is that adding this complicates the Max ALSA setting for the majority of users that prolly won't need it. Another approach would be to test in all the files for ((HDMI or Headphone) and (Software or Disabled)) and then use 0dB instead of 100%..

JWahle:
set_volume_in_db_with_select

What do you think of this UI?
Setting 100% is as easy as before and one can even detect the 4dB error with HDMI output.
This is not possible with my first proposal.

moodeaudio:
I would leave the UI alone and do the detection under the covers so user just sets 100% in the one field and the code determines whether to do 100% or 0dB. Which one is used could be explained in the help text.

JWahle:
That would also be fine with me - and probably prevents someone else from having this problem.
So, what logic would you use?
Currently, on the HDMI port the volumes are mapped: 100% = 4.00dB, 99% = 2.94dB, 98% = 1.88dB, 97% = 0.81dB, 96% = -0.25dB
If only 100% is set to 0dB, then 99% is louder than 100%.
Or should no volume above 0dB be possible? (100% = 99% = 98% = 97% = 0dB)

moodeaudio:
I think a test in all the files that set 0dB for ( (HDMI or Headphone) and (Software or Disabled) ) and then use "0dB" instead of "100%" in the amixer command.

It's an edge case though cos most users are into HAT or USB Audio devices and don't use the on-board HDMI/Headphone ports. It might be better to lobby the Pi Foundation to ditch to +4dB boost or at least make it optional.

monkeypox:
Instead of two input fields why not just use one and then compute and show the other value for people that want that number, e.g. for 100% show 0dB or vice versa if you think dB is the more important thing to show.

moodeaudio:
Its not really about showing a particular scale for that setting. The on-board HDMI and Headphone ports are configured by RaspiOS/Kernel to be ALSA 100% = +4dB instead of = 0dB. Typically one would just back off ALSA max volume to 96% or something like that to get to 0dB but the percentage scale doesn't allow fractions and none of the whole number values yield exactly 0dB.

Apparently though there is a way to get exact 0dB by using dB scale in the amixer command.
```
amixer sset Digital 0dB
```

This sets HDMI or Headphone port to exactly 0dB. If the -- "no more arguments" switch is used the command will accept negative dB values which in theory would allow using a pure dB scale for example 0 to -127dB. This would be ideal but prolly a fair amount of work to reflect through the UI and knob especially if the device's dB range had to be considered for example some devices only go to -96dB or -115dB
```
amixer -- sset Master -82dB
```

monkeypox:
Reading this on GitHub to get context I agree with Tim about just doing the right thing behind the scenes wrt 0 dB and filing an issue with the Pi foundation about the +4 dB issue with headphones and the hdmi port.

JWahle:
I reverted to the old UI and added the info text.
`util.sh` now contains an option `set-alsavol-to-max`, which implements the workaround.
It is only effective through the Max ALSA Volume setting - not when moOde volume control is used to change ALSA volume (MPD Settings: Mixer Type = "Hardware").
I also refactored all placed that called `util.sh set-alsavol ...` to use `set-alsavol-to-max`.
If you are concerned regarding the `set-alsavol` refactoring, or want the fix to apply with `Mixer Type = Hardware`, I can also fix that.

JWahle:
I just tested this with the headphone output, but it is not affected. 100% or 4dB output volume does not distort, so I excluded it from this patch.

moodeaudio:
Since ALSA volume is set by util.sh I think that might be the only file that needs to be modified. The code block that sets ALSA volume should do the test below. It just need to get some sql values. This way we can avoid another setting on System config and just explain in the help test what happens when the mixer is HDMI or Headphones.

```
if $3 == 100 and (HDMI or Headphone) and (Software or Disabled) then "0dB" else "100%"
```

```
pi@rp2:~ $ moodeutl -q "select * from cfg_system where param='amixname'"
36|amixname|HDMI

pi@rp2:~ $ moodeutl -q "select * from cfg_mpd where param='mixer_type'"
22|mixer_type|hardware
```

JWahle:
With "I reverted to the old UI" I meant, that I reverted to the vanilla moOde UI and did pretty much what you said.
There is no additional setting in the system config.
I rebased my commits, to remove some unnecessary changes from the diff.
This is what I changed in the individual files:
File | Changes
----|----
snd-config.html | Added the info, that 100% on HDMI is replaced with 0dB and other values are unchanged.
snd-config.php | Some variable names for the template were wrong - I fixed that.
util.sh | Implemented the suggested changes through the parameter `set-alsavol-to-max`
other *.sh files | Every file contained duplicated code, that loaded `alsavol_max` and `amixname` from the db<br>and call `util.sh amixname alsavol_max`.<br>I replaced all that with just `util.sh set-alsavol-to-max`.
other .php files | I replaced these calls:<br>`sysCmd('/var/www/command/util.sh set-alsavol ' . '"' . $_SESSION['amixname']  . '" ' . $_SESSION['alsavolume_max']);`<br>with `sysCmd('/var/www/command/util.sh set-alsavol-to-max');`

moodeaudio:
I'm still not following the changes, especially the introduction of get/set-alsavolume-db in util.sh. I though the only concern was 100% causes +4dB on HDMI or Headphone which can be remedied by using "0dB" instead of "100%" in the amixer command. I think this can be addressed by some mods to just util.sh.

moodeaudio:
I just committed an update that addresses the HDM/Headphone 0dB issue.
https://github.com/moode-player/moode/commit/fc894beccf9cc58bea10d0470a87c6bddbd1fb01

Please review. If this takes of the issue then go ahead and close this PR. If you would like to propose a general method of using dB for ALSA volume then feel free to open an issue so we can discuss.

JWahle:
With one small exception, the commit seems fine to me: +4dB on Headphone output is ok and does not distort.
See [here](https://github.com/moode-player/moode/commit/fc894beccf9cc58bea10d0470a87c6bddbd1fb01#r45892700)

moodeaudio:
Ok got it. I'll push an update in a bit.

JWahle:
I tested this and it works as I expected, thanks.

PR #299, opened 2021-01-09
Song lyrics

Stephanowicz:
**Search for lyrics of current playing song**

I had to create a new pull request as things got a bit messed up :D sorry...

Added a lyrics-query for the current playing song.  (You may also take a look [here](https://github.com/moode-player/moode/issues/280))
(Radio stations haven't been taken in consideration - maybe this needs additional steps.)

Requirements:

- Genius client access token
- php-dom (`sudo apt-get install php7.3-xml`)

The query is using the [genius.com api](https://docs.genius.com/#/getting-started-h1) - in order to use it You'll need to create a [free account](https://genius.com/signup_or_login). Then You need to [create a new api-client](https://genius.com/api-clients) in order to **generate a client access token**.
At moment, this token has to be put in the first line of the ./command/geniuslyrics.php file. 

The modal lyrics-window can be accessed by the playbar-menu.

Todo:

- An option to enter the token on the settings page (maybe with a small howto and a link for account creation and creation of api-clients).

image
image

 The title of the audio may differ a bit from the title in genius - In this case the query returns no result, but You can try again by changing the search string:

image

Cheers, Stephan

### DIFF
--- a/www/command/geniuslyrics.php
+++ b/www/command/geniuslyrics.php
@@ -1,0 +1,252 @@
+<?php
+/*!
+ * moOde audio player (C) 2014 Tim Curtis
+ * http://moodeaudio.org
+ *
+ * (C) 2020 Stephanowicz
+ * https://github.com/Stephanowicz
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * This Program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * 2020-12-15 TC moOde 7.0.0
+ *
+ */
+ error_reporting(E_ALL & ~E_NOTICE & ~E_WARNING);
+$ClientAccessToken = '<<<GENIUS.COM --- CLIENT ACCESS TOKEN>>>';
+
+if(count($argv)>1){
+  parse_str(implode('&', array_slice($argv, 1)), $_GET);
+}
+if(isset($_GET['artist'])){
+  $ARTIST=$_GET['artist'];
+}
+else{
+  $ARTIST_=shell_exec('mpc --format %artist% | head -n 1');
+  $ARTIST_=str_replace("\n","",$ARTIST_);
+  $ARTIST=str_replace(" ","+",$ARTIST_);
+}
+if(isset($_GET['title'])){
+  $TITLE=$_GET['title'];
+}
+else{
+  $TITLE_=shell_exec('mpc --format %title% | head -n 1');
+  $TITLE_=str_replace("\n","",$TITLE_);
+  $TITLE=str_replace(" ","+",$TITLE_);
+}
+//******************************************************************************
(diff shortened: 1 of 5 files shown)


moodeaudio:
Looking good :-) 

I'll plan to merge and test tomorrow.

monkeypox:
Maybe this could automatically run when a song plays and we could additionally fetch cover art to sub in if it’s missing? If php makes the timing tricky there’s an mit licensed js library that does a similar thing:

https://github.com/farshed/genius-lyrics-api/blob/master/README.md

It looks like it returns a promise with the relevant data, in the .then() function moode could insert the lyrics data and check the current url to see if it’s the same as the default image and if so swap in the genius url.

moodeaudio:
An auto-display option might be cool but there would need to be a way to display the lyrics w/o obscuring the whole UI with the modal.

moodeaudio:
After reviewing the TOS and some posts in their Forum it appears that they do not allow 3rd party apps to display their lyrics unless they expressly grant a license to do so. 

Forum post by Nightpool https://genius.com/nightpool
https://genius.com/discussions/371251-Genius-based-lyrics-plugin-for-winamp
```
"Unfortunately, we don’t support allowing third parties to display our lyrics without licensing.
There are legal barriers that prevent us from allowing others to display our lyrics in third party apps."
```

This means that it will not be possible to include this PR unless something changes and there is a way to obtain an app license  or they provide a general license that allows FOSS apps to use their service. If either of these things happen I can reopen the PR.

Stephanowicz:
Hmm :(

ok

bitkeeper:
@Stephanowicz too bad about the licensing, it was a nice addition.

moodeaudio:
I agree its a shame that we can't merge this nice contribution at this time but I never loose hope. Maybe someone should contact them and see whether there is an exception in their terms for FOSS (non commercial) applications that only need the capability for users too display the lyrics.

Stephanowicz:
Thank you :)

well, at least I can enjoy it. xD

Honestly, I have not much hope that genius would agree to this, having in mind that post in their forum where on guy asked for integration in winamp...

So far, cheers. Stephan

Stephanowicz:
Just came to my mind...how about displaying local stored lyrics if available? One could seperate the creation of the files from moode. So people may create them on their own or let  create  them by a script like we have here... so it's up to the user if he wants to use such a script independently from moode.

moodeaudio:
Sure if the user modifies moOde code for their own purpose then there is no liability to the project. Maybe some sort of plugin system could be created.

Stephanowicz:
how about offering options in the settings page - like 'local lyrics files' (maybe default setting) or 'external program'  (linking to a php script) ?

moodeaudio:
That might be an option if there is a viable utility that users could use to easily batch create the local lyrics files otherwise it would be a huge burden to manually download lyrics file for each track :-0

Stephanowicz:
Yes - well, as said before, the default action should be to check if there's a local lyrics file available. Otherwise, if the user has supplied a script, it will be called with the search params. The script will send back the lyrics to the ui. In addition the lyrics may be saved to a file either by the script or the ui.
So it's the liability of the user to supply a script that executes a lyrics query. 
But the process should be automated by the ui - so it's either an option that one can set in the settings or as alternative the ui may check a special folder for a script.
The interface would be as simple as possible - just artist and title as params and the plain lyrics in return.
I could imagine of an additional option to provide an own template for the script.

So the easiest option would be a special folder, e.g. "extensions/lyrics". This folder should contain a script which is able to perform a query independently - like php, perl or python. 
The interface would send artist and title as params and expects text in return.

moodeaudio:
It's not likely that many users will be developing their own lyrics downloading script. A feature like this really needs a a structured plugin system that would allow a dev to create the feature in GitHub to be installed by user.

PR #312, opened 2021-01-19
Autodetect support smb protocols

bitkeeper:
CIFS v1 isn't used by modern systems anymore, while moOde default use v1.
To improve out of the box behaviour of moOde switch the default from v1 to v2.

### DIFF
--- a/www/inc/playerlib.php
+++ b/www/inc/playerlib.php
@@ -2210,7 +2210,14 @@ function sourceMount($action, $id = '') {
 			// cifs and nfs
 			if ($mp[0]['type'] != 'upnp') {
 				if ($mp[0]['type'] == 'cifs') {
-					$mountstr = "mount -t cifs \"//" . $mp[0]['address'] . "/" . $mp[0]['remotedir'] . "\" -o username=\"" . $mp[0]['username'] . "\",password=\"" . $mp[0]['password'] . "\",rsize=" . $mp[0]['rsize'] . ",wsize=" . $mp[0]['wsize'] . ",iocharset=" . $mp[0]['charset'] . "," . $mp[0]['options'] . " \"/mnt/NAS/" . $mp[0]['name'] . "\"";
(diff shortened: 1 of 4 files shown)


moodeaudio:
We could do this but I'd prefer trying to detect the SMB version of the remote host and then automatically insert the reported version into the mount flags. If more than one version is reported then prolly default to using the highest.

Maybe something like below.

```
sudo apt-get install nmap

# RT-N66U (192.168.1.1) is host running Samba server
pi@rp2:~ $ nmap RT-N66U --script smb-protocols | grep SMBv
|_    NT LM 0.12 (SMBv1) [dangerous, but default]

```

bitkeeper:
That is even better, as long as the smb 1 is gone! It unsafe and default disabled on a lot devices.
Not only in the moOde forums but also in other forums I see people having issues with mounting smb and using the options seems too terrifying for a lot of people. 

Install of nmap will cost about 29MB.
I would suggest that settings vers at options still can be used to override the detection. But default it shouldn't be present anymore.
For this I will modify the PR to remove the defaults for vers at all.

If you change the grep filter to '|' it will show all protocols:
```
pi@moodep4:/tmp $ nmap bitspot --script smb-protocols |grep \|
| smb-protocols:
|   dialects:
|     NT LM 0.12 (SMBv1) [dangerous, but default]
|     2.02
|     2.10
|     3.00
|     3.02
|_    3.11
```

bitkeeper:
Updated name PR and added WIP flag to follow the direction of the autodetect protocol version.

bitkeeper:
@moodeaudio If you at the latest commit (still have to finishs 2 todo in the code), but is this the way you meant ?

moodeaudio:
Yes, this looks good. My Samba server runs only v1 so we can use it to test that case.

bitkeeper:
Ok done. 

Btw the detection will cost arround +/- 1 second.

PR #314, opened 2021-01-21
Browse to folder

Stephanowicz:
added option "Browse to folder" in context menu of playlist-item and playbar menu
--> Jump to the corresponding folder containing the playlist-song in folder view

image

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -2065,6 +2065,36 @@ $('.context-menu a').click(function(e) {
     else if ($(this).data('cmd') == 'track_info_folder') {
         audioinfo('track_info', path);
 	}
+  else if ($(this).data('cmd') == 'track_browse_to_folder') {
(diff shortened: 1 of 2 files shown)


moodeaudio:
The challenge with this one is that Tag and Radio views are left out. A more generalized version would be something like "Browse to Library" and would be context aware. If the PL item were a radio station it would switch to Radio view and highlight the station. If the PL item is a song file then it would switch to either Tag or Folder view depending on a Prefs setting.

Stephanowicz:
Ok, You're right - I'm too fixed on my folder jail :D

> If the PL item is a song file then it would switch to either Tag or Folder view depending on a Prefs setting.
on the other hand one would have the option to decide - well I see, it's a bit difficult with the personal preferences :)

Stephanowicz:
I will close this for now and look how these diffrent scenarios can be solved

Cheers, Stephan

PR #315, opened 2021-01-21
Track info for playlist-item

Stephanowicz:
Added an option 'Track info' for playlistitem-contextmenu

image

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -2066,15 +2066,20 @@ $('.context-menu a').click(function(e) {
         audioinfo('track_info', path);
 	}
   else if ($(this).data('cmd') == 'track_info_pl') {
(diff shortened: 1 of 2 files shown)


moodeaudio:
This is an excellent idea :-) Does it handle both track and radio station?

It looks like its intermingled with the browse to folder changes. I think it should be stand-alone.

Stephanowicz:
> This is an excellent idea :-) Does it handle both track and radio station?
> 
> It looks like its intermingled with the browse to folder changes. I think it should be stand-alone.
No radio at moment - but I will look into it - shouldn't be a problem as this is already handled elsewhere I think

Yes, the screenshot is missleading - this pr just has the trackinfo changes

Stephanowicz:
added Radiostations

PR #316, opened 2021-01-21
Albumart for multiple images

Stephanowicz:
Extracting image-array of mediatags with [getID3](https://github.com/JamesHeinrich/getID3).
If multiple images are found they will be displayed as thumbnails besides the album cover.
By clicking on them a modal window is opened with a simple image-slider.

Requires: [getID3](https://github.com/JamesHeinrich/getID3)
getID3 is expected to be located in inc/getid3

image

image

### DIFF
--- a/www/css/moode.css
+++ b/www/css/moode.css
@@ -562,6 +562,22 @@ ol.playhistory {margin-left:55px;}
 #ph-filter-results {line-height:28px;font-size:14px;font-style:italic;}
 #ph-search {margin-bottom:8px;}
 #ph-search input {width:145px;margin:0;padding: 2px 0 2px 10px;border-top-right-radius:0px;border-bottom-right-radius:0px;}
+/* Albumart.html */
+#slider{opacity:1;transition:opacity 0.25s;}
+#slider.fadeOut{opacity:0;}
+.dot {
+cursor: pointer;
+height: 15px;
+width: 15px;
+margin: 0 2px;
+background-color: #999999;
+border-radius: 50%;
+display: inline-block;
+transition: background-color 0.3s ease;
+}
+.dotactive, .dot:hover {
+background-color: #111111;
+}
 
 /* Use custom scroll bars for desktop monitors */
 @media (pointer: fine) {
--- a/www/footer.php
+++ b/www/footer.php
@@ -170,6 +170,9 @@
 	<span class="reconnect-msg">System shut down</span>
 </div>
 
+<!-- AlbumArt -->
+<div id="albumart-modal" class="modal hide" tabindex="-1" role="dialog" aria-labelledby="albumart-modal-label" aria-hidden="true"></div>
+
 <?php
     //workerLog('-- footer.php');
     $return_val = session_write_close();
--- a/www/getImageID3.php
+++ b/www/getImageID3.php
@@ -1,0 +1,73 @@
+<?php
+/**
+ * moOde audio player (C) 2014 Tim Curtis
+ * http://moodeaudio.org
+ *
+ * (C) 2021 Stephanowicz
+ * https://github.com/Stephanowicz/moode/
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * This Program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * 2021-MM-DD TC moOde 7.x.x
+ *
+ */
+
+require_once dirname(__FILE__) . '/inc/getid3/getid3.php';
+$file=shell_exec('mpc -f %file%|head -n 1');
+if($file){
+  $file=trim ($file);
+  $path=shell_exec('sudo egrep \'^music_directory\' /etc/mpd.conf | grep -Po \'(?<=")[^"]*\'');
+  $path=trim ($path);
+  $path.='/'.$file;
+  $img = getImage($path);
+  if($img){
+    echo json_encode($img);
+  }
+  else {
+    echo "";
+  }
+  die;
+}
+
+function getImage($filename)
+ {
+     if (!file_exists($filename)) {
+         return null;
+     }
+     $getID3 = new getID3();
+     // Analyze file and store returned data in $file_info
+     $file_info = $getID3->analyze($filename);
+     // extract the album artwork
+     $artwork = null;
+     //flac ogg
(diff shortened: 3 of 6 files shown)


moodeaudio:
Great idea :-) but I have some questions.

1. Is the cover art layout with the thumbs conditional? In other words do the thumbs only appear if multiple cover images exist?
2. Are cover files handled or is it only embedded images?
3. Which audio formats are supported?

monkeypox:
Could you post a bigger image of the screen with the thumbs? It looks like the thumbs shift the cover over which seems like it’d make things seem unbalanced for the majority of time you probably don’t care about additional images, but it’s hard to tell without the context of the rest of the screen. If nothing else it’d be nice to have some padding between them.

Alternately if just allowing images to be shown is the important part rather than showing available images in thumbs you could do dots under the image (if there are extra images) like on your modal and let users swipe/click between them. I have a better swipe function I’ll check in one of these days that isn’t so resource hoggy.

Stephanowicz:
> Great idea :-) but I have some questions.
> 
> 1. Is the cover art layout with the thumbs conditional? In other words do the thumbs only appear if multiple cover images exist?
> 2. Are cover files handled or is it only embedded images?
> 3. Which audio formats are supported?

1. At moment thumbs will be displayed if one is available - I'm undecided as both options are nice but I'm tending more to only showing them if more than one is available
`if(Array.isArray(imgArray) && imgArray.length > 0)`  --> `imgArray.length > 1`

2. It's handling only embedded images. But it may be enhanced in future to provide different sources of images...

3. Well, that's a long list... please take a look here: [https://www.getid3.org/](https://www.getid3.org/)

Stephanowicz:
> Could you post a bigger image of the screen with the thumbs? It looks like the thumbs shift the cover over which seems like it’d make things seem unbalanced for the majority of time you probably don’t care about additional images, but it’s hard to tell without the context of the rest of the screen. If nothing else it’d be nice to have some padding between them. Alternately if just allowing images to be shown is the important part rather than showing available images in thumbs you could do dots under the image (if there are extra images) like on your modal and let users swipe/click between them. I have a better swipe function I’ll check in one of these days that isn’t so resource hoggy.

Yes, padding is an idea.
At moment this is an additional component which does not interfere with the present albumart.

**So a different concept would be** to show indicators like dots below the cover if more images are available and the possibility to swipe or tap or whatever. But to get the modal/fullscreen view an additional option is needed - maybe like showing the thumbs in the audioinfo page or a fullscreen icon next to the dots...

image

Stephanowicz:
>  I have a better swipe function I’ll check in one of these days that isn’t so resource hoggy.

Do You mean to load the images on demand?
At moment they are stored in the browser cache as variable.

monkeypox:
> Yes, padding is an idea.

Seeing it full screen was great thank you. A little vertical padding between thumbs would look nicer IMO, and also having symmetry between the screen edge <-> thumbs <-> cover image. 

> At moment this is an additional component which does not interfere with the present albumart.

As an aside looking at their project page the ability to write mp3 tags is interesting and would help differentiate moode, lord knows it could solve a lot of hassles with badly tagged files but it’s probably also a big project.

> **So a different concept would be** to show indicators like dots below the cover if more images are available and the possibility to swipe or tap or whatever. But to get the modal/fullscreen view an additional option is needed - maybe like showing the thumbs in the audioinfo page or a fullscreen icon next to the dots...

You could also have a swipe / button to show the thumbs if available.

Also:

1) how would this look on mobile / portrait?

2) the additional images should probably also be available in cover view.

3) is there a limit on the number of images?

4) if the modal doubles as a full screen zoom of the cover you could have thumbs present there (or as well), there should be plenty of either vertical or horizontal room to spare depending on the device screen orientation. 

5) people will want support for non-embedded images as well so we should look at that too, at least as a future option.

6) there’s a cover hash moode checks to avoid reloading the cover art so you could add a call to do the image fetching based on that (if you’re not already) to minimize how often it has to run this though that would limit the ability to have per track images if that’s a thing.

Stephanowicz:
> > Yes, padding is an idea.
> 
> Seeing it full screen was great thank you. A little vertical padding between thumbs would look nicer IMO, and also having symmetry between the screen edge <-> thumbs <-> cover image.
> 
> > At moment this is an additional component which does not interfere with the present albumart.
> 
> As an aside looking at their project page the ability to write mp3 tags is interesting and would help differentiate moode, lord knows it could solve a lot of hassles with badly tagged files but it’s probably also a big project.

Yes - a tag-editor would be nice :D

> > **So a different concept would be** to show indicators like dots below the cover if more images are available and the possibility to swipe or tap or whatever. But to get the modal/fullscreen view an additional option is needed - maybe like showing the thumbs in the audioinfo page or a fullscreen icon next to the dots...
> 
> You could also have a swipe / button to show the thumbs if available.
> 
> Also:
> 
> 1. how would this look on mobile / portrait?

this might be a problem :(  or better: it is a problem - so thumbs should only show up if enough space is available
then switching to dots below if space is too small

> 2. the additional images should probably also be available in cover view.

I'll look into that

> 3. is there a limit on the number of images?

I did not limit it - but maybe a good idea to allow only a certain amount regarding loading time

> 4. if the modal doubles as a full screen zoom of the cover you could have thumbs present there (or as well), there should be plenty of either vertical or horizontal room to spare depending on the device screen orientation.

You mean You would prefer thumbs to the dots?

> 5. people will want support for non-embedded images as well so we should look at that too, at least as a future option.

yes - not now

> 6. there’s a cover hash moode checks to avoid reloading the cover art so you could add a call to do the image fetching based on that (if you’re not already) to minimize how often it has to run this though that would limit the ability to have per track images if that’s a thing.

yes, I put it inside this code - so it will only be called when the hash has changed - that should be enough

monkeypox:
> > 1. how would this look on mobile / portrait?
> 
> this might be a problem :( or better: it is a problem - so thumbs should only show up if enough space is available
> then switching to dots below if space is too small

I don’t have any music this would apply to so it doesn’t affect me but my concern is if implemented as-is people would want the option to turn it off and adding more toggles increases the cognitive load of configuring moode. As much as possible we should have options like this just be built into the interface rather than controlled via preferences. The dots under the cover that you swipe / click to see are pretty unobtrusive so I don’t think need a toggle, ditto a <-> kind of full-screen icon (which might also be useful for a regular cover for people with high-res cover art). You could also have a way to toggle the thumbs display, it could still be stored as a config option so the change was sticky.

> > 1. if the modal doubles as a full screen zoom of the cover you could have thumbs present there (or as well), there should be plenty of either vertical or horizontal room to spare depending on the device screen orientation.
> 
> You mean You would prefer thumbs to the dots?

Thumbs seem like they might be nicer if you’re in the modal and have more room to display them. It might also be beneficial from a memory standpoint to load an html template into a “empty” div like we do with the track/hardware dialog for the modal but I haven’t looked to see if that’s workable here.

moodeaudio:
My concern is that AFAIK most user collections don't have multi-art and so I'm not getting the vibe that this is something that is generally applicable.

Stephanowicz:
> My concern is that AFAIK most user collections don't have multi-art and so I'm not getting the vibe that this is something that is generally applicable.

well - @moment thumbs will only be displayed if more than one image is available 
I will close this now for the moment and think about it and test different approaches.

Cheers, Stephan

PR #321, opened 2021-01-26
Default hide some camilladsp settings.

bitkeeper:
Default the settings are hiden, but can be enabled again by toggling the expert mode to on:
image

Without hide of the settings:
image

With hide of the settings:
image

Requires new camillagui & camillagui-backend to make the it work, but will not break the current camillagui & camillagui-backend.

### DIFF
--- a/other/camilladsp/gui/_install.sh
+++ b/other/camilladsp/gui/_install.sh
@@ -19,6 +19,7 @@ sudo mkdir -p /opt
 sudo unzip -o camillagui-v$CAMILLAGUI_VER.zip -d /opt/camillagui
 
 sudo cp camillagui.yml /opt/camillagui/config
+sudo cp gui-config.yml /opt/camillagui/config
 sudo cp camillagui.service /etc/systemd/system
 
 # Don't start it, start it when needed from the moode config pages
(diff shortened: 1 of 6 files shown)


bitkeeper:
@JWahle support for hide of the settings is added.
Run your own version of camilla-gui & backend version, until we known if we need to roll our own or a new release of teh camillgui is available in time.

JWahle:
I think with the exception of the capture device all other options should always be hidden.
"hide_capture_device" should be true, if and only if your switch to automatically replace the output device in the config is true (if you did not change/remove it in the mean time).

`capture_samplerate` is always overridden by ALSA CDSP - no need for it.
`Silence` is only needed to stop processing, when CDSP is running as a service. moOde automatically stops CDSP when paused/stopped, so no need for it either.
`Capture device` is always Stdin (or File "/dev/stdin"), channels and format are always overridden by ALSA CDSP.
With the audio sources of moOde I don't see any reason to set `read_bytes` or `skip_bytes`.
Now, that I think of it, the `extra_samples` option should probably be visible, even if the capture device is hidden - I'll fix that.

JWahle:
I started to implement an option to show/hide `extra_samples` independently of the other properties, but is was more fiddly than I liked, so I stopped. Not sure, if it is really that important. If the need arises, I might do it later.

bitkeeper:
@JWahle I'm very happy with your current implementation, the camillagui now feeld decent integrated.
So yes lets leave things as they are now.

PR #322, opened 2021-01-26
Multiple albumart

Stephanowicz:
Hello again,

so, I made some polishing and improvements to the code and hope the result is satisfying :)

requirements: 

- [getID3](https://github.com/JamesHeinrich/getID3) lib  (for supported audiofiletypes take a look [here](https://www.getid3.org/))
- the getID3 lib folder is expected to be located at inc/getID3

This code is an additional component which does not interfere with the present albumart code.
The function is triggerd whenever the hash of the last cover has changed.
At moment only embedded images are processed. But it may be enhanced in future to provide different sources of images.
If more than one embedded image is available, the images will be shown as thumbnails next to the albumcover.
Also dot indicators are shown below the albumcover. When clicking on them the albumcover changes accordingly.
By clicking a thumb image or the expand symbol next to the dots, an overlay will open with a maxed view of the images.

The thumbnails will get hidden when the screenwidth is below 780px - the dots will still be visible
The dots and the expand symbol are also available in the coverview/screensaver

image

image
[modal view]

image
[cover view / screensaver]

image
[screen width < 780px]


monkeypox:
1) have you thought about overlaying the zoom button onto like the lower right of the album cover to allow the dots to always be centered?

2) why have both dots and thumbs at the same time? unless you can toggle thumbs off at which point I think you’d want to toggle one or the other off. 

3) the dots seem too big in context on the playback screen, half that diameter is probably closer to the mark - visible but not distracting.

Stephanowicz:
> 1. have you thought about overlaying the zoom button onto like the lower right of the album cover to allow the dots to always be centered?

Yes - that's a bit of a 'problem' - I didn't want to intervent to much into the existing code. 
It would be possible to adjust by putting the cover, the dots and the thumbs into a table - I will check this

> 2. why have both dots and thumbs at the same time? unless you can toggle thumbs off at which point I think you’d want to toggle one or the other off.

The thumbs open the modal view
The dots switch through the images in the main view

> 3. the dots seem too big in context on the playback screen, half that diameter is probably closer to the mark - visible but not distracting.

Yes, but then it's hard to click them.
one could put other controls like arrow buttons hovering over the cover for the navigation - but this might be a bit difficult as the image below is linked to open the lib-view

monkeypox:
> > 1. have you thought about overlaying the zoom button onto like the lower right of the album cover to allow the dots to always be centered?
> 
> Yes - that's a bit of a 'problem' - I didn't want to intervent to much into the existing code.
> It would be possible to adjust by putting the cover, the dots and the thumbs into a table - I will check this

You should use css and absolute positioning to position the element. 

> > 1. why have both dots and thumbs at the same time? unless you can toggle thumbs off at which point I think you’d want to toggle one or the other off.
> 
> The thumbs open the modal view
> The dots switch through the images in the main view
> 
> > 1. the dots seem too big in context on the playback screen, half that diameter is probably closer to the mark - visible but not distracting.
> 
> Yes, but then it's hard to click them.
> one could put other controls like arrow buttons hovering over the cover for the navigation - but this might be a bit difficult as the image below is linked to open the lib-view

Wouldn’t people expect to use the thumbs to change to the different images?

Anyway if Tim wants to check this in I think we should do something like:

1) have the thumbs toggle on/off via a control on the playback screen, save this state to the db so it’s sticky. I’m not sure what form this should take, maybe something overlaid on the cover.

2) on mobile portrait only show dots & add support to swipe between them, the image shown is already essentially “full screen” anyway so no need for the modal.

3) thumbs control the image shown on desktop / landscape, no dots. we could add swipe between for touch devices but to make things simple maybe not.

4) use a separate control to open the modal view, probably in-line with the thumbs, but maybe we don’t really need a modal view.

I can help on the css and layout if it isn’t something you are comfortable doing but be aware I don’t have a lot of time for moode work right now.

moodeaudio:
The challenge is that multi-art collections are an edge case. Whats the justification for doing this?

Stephanowicz:
> The challenge is that multi-art collections are an edge case. Whats the justification for doing this?

Ok - well, maybe be it's an edge case because it's not well supported in mediaplayers? :D 
I've been using [musicbee](https://getmusicbee.com/) before I switched to rapsi and  moode. With mb I first realized the possibility of this and then started to use it whenever possible.

So, do You mean this in general regarding multiple images or the specific case of embedded images?

moodeaudio:
You are prolly right that the reason we don't see many multi-art collections is that most players don't support multiple cover art images per album but also I think most free metadata repositories only maintain the front cover and similarly most tag/rip utilities only support one cover.

Yes, I think it's generally an edge case but we've done a lot of edge case coding in moOde so thats not a deal breaker and I'm sure the users that have asked for this in the past would be happy to see it although IIRC their collections had cover art images in the album dir rather than embedded in the tracks.

So let's say for the purpose of discussion that we agree that this should be done. The question is then how to implement it. One thing that is somewhat sacred in moOde is the layout and minimalism of the views, and particularly Playback view. 

Generally if a UI element or function is frequently used then it prolly needs to be always visible otherwise it should be on a menu or a config screen. II I apply that line of thinking to the multi-art case it would suggest that it should be part of something like the "Track info" screen.

Since your earlier PR added "Track info" selection to the PL context menu and we already have it on the context menus for Tag and Album views, users would be able to see their covers and all the metadata in any view.

monkeypox:
A bonus of the modal only method is mobile and desktop layout/css should be basically the same so it’d reduce complexity.

As a compromise to improve discoverability on albums with multiple covers it wouldn’t add much clutter to overlay a button onto a bottom corner of the cover art of albums that have multiple covers like we discussed earlier. Maybe one of the FA circle icons like caret-circle-right?

https://fontawesome.com/icons/caret-circle-right?style=regular

100% white should be visible on virtually all covers.

Stephanowicz:
> So let's say for the purpose of discussion that we agree that this should be done. The question is then how to implement it. One thing that is somewhat sacred in moOde is the layout and minimalism of the views, and particularly Playback view.

Ok - maybe I was a bit too bold with storming in with my ideas like that :D  
You spent much time and effort in developing and now a newbie pushes in and wants to change everything :D - sorry.

 

> II I apply that line of thinking to the multi-art case it would suggest that it should be part of something like the "Track info" screen.

Well, sure - though I'd prefer to have it on the main screen - the Trackinfo is also a sensible place.
(an option in the preferences menu would also be a solution, so the user can decide if he wants this in the playback view...)

> although IIRC their collections had cover art images in the album dir rather than embedded in the tracks.

the problem I see with Folder pictures is that there might be a lot of 'rubbish' - well, one could differentiate them by a minimum size they need to have to be selected for display at least.

Ok, so I will have a look to add this to the trackinfo 

Cheers, Stephan

moodeaudio:
lol, keep the storm coming! Thats how ideas get refined into great features and there is nothing better than to see actual code  and a mockup to stimulate a discussion :-)

As far as cover files go I remember from a few years back that the chat was to have a gallery or lightbox type viewer that would display any images (or pdf's if possible) that were found in the dir. Maybe check the Pref setting "Cover search priority " to determine whether to check for embedded first or cover files first. Something like that.

Stephanowicz:
Ok - I can't help it! Next storm-warning :D

So, what do You think about this - it's a kind of compromise, or a win-wing thing:

- I removed the dots and the expand icon. 
- The thumbnails will become visible when hovering over the albumcover-area - and hide when leaving it.
- Clicking on a thumb changes the cover - doubleclicking it opens the fullscreen modal.
- Also tried it with my touch display - clicking in the albumart-area (outside of any links) also shows the thumbs 
- resized the thumbs to 30px from 50px


image

image

moodeaudio:
I think a multi-cover feature has to handle both embedded covers and cover files same as the current single-cover function because users are going to populate their collections both ways.

The main challenge with putting thumbs directly on Playback view is that the covers other than the front cover are not something that will be frequently viewed and it clutters up the view.

I think Track Info would be a better place for the covers. It's easily available in all the views and it can be very detailed. An example mockup could place the cover thumbs in a row between the [Track] [Hardware] buttons and the start of the metadata.

Stephanowicz:
Hi Tim,

yes - I got that with the 'sacred' view :D and I think I understand the reservations...

As I wrote in my last post, the thumbs will only be visible when hovering over the cover area (and only if there are more than one) - they will fade in while the main cover's size will be reduced smoothly and vice versa when leaving the area.
(I've also been thinkin of a small indicator as an alternative to the hovering somewhere next to the cover - so that the thumbs will only show when clicking on it instead of hovering, if hovering is also too intrusive...)

I would see this as a first step - and people may evalute this 
A next step would be to integrate files in the folder

Maybe this could be offered as an option in the preferences - so people can decide

The Trackinfo is definetly a perfect place - in addition :D  and I will tackle that next

moodeaudio:
I would sequence it by integrating into Track info first including handling cover files and then see how that goes. The advantage is that it is available in all views. If there ends up being user demand for doing something directly in Playback view we can see whether it makes sense.

Stephanowicz:
Yep, makes sense 

Cheers, Stephan

Stephanowicz:
PS: Just was wondering about ALAC support of getID3 as I recently read that ALAC has become open source long time ago... 
So just downloaded a sample and tested it with getID3 - no probs! :D

PR #326, opened 2021-01-30
Add stationmanager.py script for radio import/export including filtering

bitkeeper:
New python script for handling import/export of radiostation backups which support filter, merging and compare.
Script doesn't require any other python that isn't already installed with moOde.

Features:
- backup backup/restore same moode version
- backup/restore from 6.x.x to 7.x.x
- backup/restore from 7.x.x to 6.x.x
- also supports the 7.1.0 scheme changes
- import of the previous moode stationbackup format
- single script no need to install other deps
- scope attribute works well with both 6.x.x and 7.x.x.
- autodetect supported fields

Usage:
```
pi@moodep4:~/moode/www/command $ sudo ./stationmanager.py -h
usage: stationmanager.py [-h] [--version]
                         (--import | --export | --clear | --compare)
                         [--scope {all,moode,other}]
                         [--type {favorite,regular,hidden,nothidden}]
                         [--how {clear,merge}] [--db DB]
                         [backupfile]

Manages import and export of moOde radiostations.

positional arguments:
  backupfile            Filename of the backupfile. Required by the import,
                        export and compare.

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --import              Import radio stations from backup.
  --export              Export radio stations to backup.
  --clear               Clear radio stations. This will delete database,
                        images and pls files of the selected stations by the
                        scope.
  --compare             Show difference between database and backup.
  --scope {all,moode,other}
                        Indicate to which stations to apply the action.
                        (default other)
  --type {favorite,regular,hidden,nothidden}
                        Indicates the type of stations should takenin
                        account(only applies to export).
  --how {clear,merge}   On import clear stations before action or merge and
                        add. (default merge)
  --db DB               File name of the database. Default uses the standard
                        moOde db.

Root privileges required for import or clear.
```

### DIFF
--- a/www/command/moode.php
+++ b/www/command/moode.php
@@ -655,11 +655,7 @@ else {
 			echo json_encode(parsePlayHist(shell_exec('cat /var/local/www/playhistory.log')));
 			break;
 		case 'export_stations':
-			syscmd('sqlite3 /var/local/www/db/moode-sqlite3.db -csv "select * from cfg_radio" > /var/local/www/db/cfg_radio.csv');
-			syscmd('sqlite3 /var/local/www/db/moode-sqlite3.db ".schema cfg_radio" > /var/local/www/db/cfg_radio.schema');
-			sysCmd('zip -q -r ' . EXPORT_DIR . '/stations.zip /var/lib/mpd/music/RADIO/* /var/local/www/imagesw/radio-logos/* /var/local/www/db/cfg_radio.csv /var/local/www/db/cfg_radio.schema');
-			syscmd('rm /var/local/www/db/cfg_radio.csv');
-			syscmd('rm /var/local/www/db/cfg_radio.schema');
+			syscmd('/var/www/command/stationmanager.py --scope all --export ' . EXPORT_DIR . '/stations.zip');
 			break;
 		case 'import_stations':
 			if (isset($_FILES['stationbackupfile'] ) == false ) {
@@ -678,9 +674,12 @@ else {
 				}
 				else {
 					// Import station data from zip
-					$result = sysCmd('/var/www/command/import_stations.sh');
-					if (!empty($result[0])) {
-						$msg = file_get_contents('/tmp/station_import_error.txt');
+					$output = array();
+					$exitcode = -1;
+					$cmd = 'sudo /var/www/command/stationmanager.py --scope all --how clear --import ' . $file;
+					exec($cmd, $output, $exitcode);
+					if($exitcode!=0) {
+						$msg= implode("\n", $output);
 						str_replace("\"", '', $msg);
 					}
 					else {
--- a/www/command/stationmanager.py
+++ b/www/command/stationmanager.py
@@ -1,0 +1,578 @@
+#!/usr/bin/python3
+#
+#  Script to for managing moOde radio stations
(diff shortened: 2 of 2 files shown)


moodeaudio:
For sure will be useful cos many users like the radio feature and want to preserve their customizations :-)

This is designed to replace the basic import/export feature I included in moOde 7 and also Kents migration scripts right?

bitkeeper:
Yes, the goals is to provide a single script that can cover the all radio backup and restore functionality, including support for older and future moOde versions and extended filtering/merging.

moodeaudio:
I'm thinking of doing the 7.1.0 release this month. Did you have some ideas on how you wanted this to be integrated into the Radio Manage?

bitkeeper:
I'm first waiting to get the blessing from Kent ;-) If everything goes as it supposed to be the script can be used on older and current versions of moOde. Script doesn't have any additional deps.
The format of the backup is different from the current one, so older backups can't be reused.

The first integration would be just replace the current implementation with this one for a full backup restore. For this one only backend changes need (calling differnt scripts)

Next would be option selective backup/restore (all|moode|user), with merge(station which already exists are ignored) or clear (delete all|moode|user stations before import mode). But this would require changes to import and export of stations.
Maybe wait with that until we maybe have a own page for backup/and restore including the system settings ?

But it isn't that important how I think about it, I'm just the guest in your house ;-), What do you think about it ?

moodeaudio:
Does the comment below mean that the zips from the current export can't be imported by the new script? I'm just wondering what we do to handle this case. Maybe check the zip for an indicator and use either the old or the new import script ?? 
```
"The format of the backup is different from the current one, so older backups can't be reused."
```

bitkeeper:
Check the zip would be easy something like:
```
unzip -Z station_import.zip var/local/www/db/cfg_radio.csv
```
and use the exit code for check.
Then it is easy to switch between old and new import script.

I will add that construction to the PR.

TheOldPresbyope:
Best hold off a bit on this. Something happened in the last couple of commits which has lead to breakage concerning the logo images.

bitkeeper:
@TheOldPresbyope can you be a bit more specific ?

Tested with the latest version (the rebased squashed one of hour ago) under:
- from 7.10pre (~branch develop) to 7.1.0pre
- from 6.5.2 to 6.5.2
- from 7.1.0pre to 6.5.2
- from 6.5.2 to 7.1.0pre

With the command:
```
sudo ./stationmanager.py --scope all --export ~/rs_all.zip 
sudo ./stationmanager.py --scope all --clear 
sudo ./stationmanager.py --scope all --how clear --import ~/rs_all.zip 
```

And I see images:
- in the radio view
- in the playlist
- and the cover in playview

The only thing I noticed is that with 6.5.2 I first have to update the library to get the radio view correct and that with 7.x the radio view is correct after a refresh.

TheOldPresbyope:
My apology for going silent---had a IRL moment.

Yes, I got caught in the same library and browser cache issues. Looks like we're good to go.

bitkeeper:
No problem Kent, we all have been there.
@moodeaudio it is ready to merge.

Maybe one of you could after merge correct some messages?

moodeaudio:
Nice :-)

I'll go through the code and clean up any messages or typos.

PR #334, opened 2021-02-10
Library option to ignore cue files

bitkeeper:
This is item that is coming back in the support forums, espcially because the behaviour of MPD change somewhere arround 0.21.25/26.  By making ignoring cuefiles default you have the old behaviour back, with the flexibilty to still include the cuefiles if you want.

image

### DIFF
--- a/var/local/www/db/moode-sqlite3.db.sql
+++ b/var/local/www/db/moode-sqlite3.db.sql
@@ -423,6 +423,7 @@ INSERT INTO cfg_system (id, param, value) VALUES (149, 'recorder_status', 'Not i
 INSERT INTO cfg_system (id, param, value) VALUES (150, 'recorder_storage', '/mnt/SDCARD');
 INSERT INTO cfg_system (id, param, value) VALUES (151, 'volume_db_display', '0');
 INSERT INTO cfg_system (id, param, value) VALUES (152, 'search_site', 'Google');
+INSERT INTO cfg_system (id, param, value) VALUES (153, 'cuefiles_ignore', '1');
 
 -- Table: cfg_theme
 CREATE TABLE cfg_theme (id INTEGER PRIMARY KEY, theme_name CHAR (32), tx_color CHAR (32), bg_color CHAR (32), mbg_color CHAR (32));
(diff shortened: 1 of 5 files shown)


moodeaudio:
IIRC MPD has always recognized cue files. There was a PR way back from @pccasto that had a set of changes to improve how cue files were handled in the Library/. Maybe we should look at that PR again??

bitkeeper:
We already have been in this dicussion see https://github.com/moode-player/moode/pull/181 ;-)
I just implemented it the toggle variant of this one.

Something I see you reaction in the forums; like nothing has changed but it has ... really.
The way mpd handles the cue vs the track has changed, from a certain version the files show double(from the cue + the file itself). If you install an older version like moode 6.5.x you will not get double tracks in your library.
You can see that in the reaction from the users; like `he is there something changed; because after upgrade I have double files.`

For we 7.1.0 we could use this one, then we at least have an option for the users. And for later see what the pccasto PR was about, ok?

moodeaudio:
lol, I can't keep track of all the discussions.

I still don't get whats happening with cue files. Normally there is a single big flac file and an associated cue file. Is it that current MPD is not reading the cue file correctly?

monkeypox:
Does mpd double the tracks when adding them through mpc or is this a moode specific thing?

moodeaudio:
I think it happens during mpd update cos moOde just parrots whats in the mpd database. I'm still a bit in the dark though cos I don't have any cue/flac test files anymore and so I can't repro the issue.

No one has mentioned it but whats the status of CUE as a format? Is it still an important format that should be  supported?

bitkeeper:
It is a MPD thing.

The CUE file isn't always about playlist list for one big flac or wav:
When ripping my cd collection Exact Audio Copy creates a log and CUE file. This CUE file it possible to recreated the CD. Besides that, a file is generated per track. That is the case I most encounter when other people get into problems with double tracks in MPD.

moodeaudio:
Ok got it. I'll merge this after I finish up another commit I'm working on.

moodeaudio:
Is there supposed to be a double splat **/ in the pattern?

```
pi@rp2:~ $ cat /var/lib/mpd/music/.mpdignore 
**/*.cue
```

bitkeeper:
Good catch, nope that should just be `*.cue`.

bitkeeper:
Fixed in PR #337

PR #338, opened 2021-02-17
[stationmanager] add support for legacy backup import of m671

bitkeeper:
Add also support for 'legacy' station backup of moOde 6.7.1. 
This one is a little bit tricky because no schema is present in those backups.

### DIFF
--- a/www/command/stationmanager.py
+++ b/www/command/stationmanager.py
@@ -37,6 +37,7 @@ from os import path
 from zipfile import ZipFile
 import sys
 import csv
+import re
 
 VERSION = "1.0"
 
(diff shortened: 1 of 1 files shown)


bitkeeper:
@TheOldPresbyope I remove the draft status after  you have verified that it works ... I hope

TheOldPresbyope:
As usual, you beat me to it :) Will test in a bit.

---

A few business rules:

There was no export function for release < 6.6.0 so only stationmanager.py or my scripts would be usable

IF legacy format AND no .schema file THEN release ( 6.6.0 | 6.7.0 | 6.7.1 )

IF legacy format AND .schema file THEN release ( 7.0.0 | 7.0.1 )

---

release < 6.6.0: fields = id, station, name. type, logo

release ( 6.6.0 | 6.7.0 | 6.7.1 ): fields = id, station, name, type, logo, genre, broadcaster, language, country, region, bitrate, format

release ( 7.0.0 | 7.0.1 ): fields = id, station, name, type, logo, genre, broadcaster, language, country, region, bitrate, format, geo_fenced, reserved1, reserved2

release 7.1.0: fields = id, station, name, type, logo, genre, broadcaster, language, country, region, bitrate, format, geo_fenced, home_page, reserved2

Note that with release 7.1.0 the "reserved1" field is now used as "home_page"

TheOldPresbyope:
First, let me apologize for my post to the Test Team that I'd identified "two problems, one in logic and one in coding...." The first stemmed from my misunderstanding of the phrase "legacy format", which I took to include previous major releases. This latest PR both makes clear it did not and fixes the problem. The second problem was real but is moot now because of changes to take care of the first.

I can now export a stations.zip file (in which I've included some "other" stations) using the Radio Manager function in moOde 6.7.1 and import it into moOde 7.0.0, 7.0.1, and 7.1.0 using the new stationmanager.py. Once the servers are restarted and the browser refreshed, all the stations are usable (although not all the URLs are currently valid, but that's a horse of another color). All the logos display except for the 320K BBC stations and for the WQXR Q2 station. I get a broken path icon for each instead. We've discussed the BBC stations before; I'll chase down the WQXR Q2 issue later. 

I've also tested using stations.zip files exported by the Radio Manager function in moOde 7.0.0 and by moOde 7.0.1, in each case importing the files into later moOde releases, include 7.1.0TR1, using the new stationmanager.py.

A secondary issue which I've noticed is that because of the way the --clear option works, orphan logo files can be left behind. Depending on the source and sink release, I've noticed the 320K BBC station logos or the Audiophile station logos left after --clear --scope all. The first relates to previous observation; the second to their "OFFLINE" status in the source release.

All in all, I'd say take this PR out of draft status so it'll get merged and others can play with it while we tinker.

bitkeeper:
You don't have to apologize for anything. Normally it is already hard enough to prevent miscommunication, let alone with working people in other timezone with another language. Will not be last time ;-)
At least we making progress.

So the second where probally multiple genres with legacy import, yes that one was fixed also.

Often the bad thing about supporting 'legacy' mode, is that new code starts to smell will all kind of construction to cover the older modes. For example the 320K BBC, te `logo != local` is supported when the new stationmanager is used to create a backup, but if the legacy import would be supported it would introduce smelly code in the new exporter.

I would advise to always make a backup onolder system with the new stationmanager.py, but that is not always feasible because the user would need to use commandline toiols

To make it easier for the user, on import gui support for scope (all | moode | other) and how (clear | merge) would be welcome.

bitkeeper:
@moodeaudio : this one is ready to merge.

PR #347, opened 2021-02-22
Fixed SoX resampling quality defaulting to 'high' for all presets.

aposcic:
SoX resampling quality was defaulting to 'high' if anything but the custom recipe option was selected.

### DIFF
--- a/www/inc/playerlib.php
+++ b/www/inc/playerlib.php
@@ -1922,7 +1922,7 @@ function updMpdConf($i2sdevice) {
 				$data .= $cfg['value'] == 'disabled' ? '' : $cfg['param'] . " \"" . $cfg['value'] . "\"\n";
 				break;
 			case 'sox_quality':
-				$sox_quality = ($cfg['value'] == 'custom' && ($patch_id & PATCH_SOX_CUSTOM_RECIPE)) ? $cfg['value'] : 'high';
+				$sox_quality = $cfg['value'];
 				break;
 			case 'sox_multithreading':
 				$sox_multithreading = $cfg['value'];


moodeaudio:
Indeed it is a bug. There is prolly a simpler way to fix it. Consider that the "Custom" option won't be offered in the list due to the logic below in mpd-config.php.

```
// Resampling quality
$_mpd_select['sox_quality'] .= "<option value=\"very high\" " . (($mpdconf['sox_quality'] == 'very high') ? "selected" : "") . " >Very high</option>\n";
$_mpd_select['sox_quality'] .= "<option value=\"high\" " . (($mpdconf['sox_quality'] == 'high') ? "selected" : "") . " >High (Default)</option>\n";
$_mpd_select['sox_quality'] .= "<option value=\"medium\" " . (($mpdconf['sox_quality'] == 'medium') ? "selected" : "") . " >Medium</option>\n";
if ($patch_id & PATCH_SOX_CUSTOM_RECIPE) {
	$_mpd_select['sox_quality'] .= "<option value=\"custom\" " . (($mpdconf['sox_quality'] == 'custom') ? "selected" : "") . " >Custom recipe</option>\n";
}

```
Therefore the fix should probably be like below.

```
			case 'sox_quality':
				$sox_quality = $cfg['value']';
				break;
```
If you agree and would like to update the PR that would be great :-)

-Tim

aposcic:
> If you agree and would like to update the PR that would be great :-)

Agreed & updated. 🙂 I was trying to stay in the spirit of the original code (with the additional guard), but this is cleaner.

Thanks, Tim!

Antonio

PR #349, opened 2021-02-24
[moodeutl] Show current audio format while device is busy

bitkeeper:
`moodeutl -f` while playing returns that the device is busy.
This PR makes it possible that also when the device is busy the audio format is returned.

In addition `moodeutl -F` will return, while the device is busy, the following:
```
pi@moodedev:/usr/share/camilladsp $ moodeutl -F
Audio device busy
Cambridge Audio Cambridge Audio USB Audio 2.0 at usb-0000:01:00.0-1.4, high spe : USB Audio

Playback:
  Status: Running
    Interface = 1
    Altset = 1
    Packet Size = 72
    Momentary freq = 44105 Hz (0x5.8360)
    Feedback Format = 16.16
  Interface 1
    Altset 1
    Format: S32_LE
    Channels: 2
    Endpoint: 1 OUT (ASYNC)
    Rates: 44100, 48000, 88200, 96000, 192000
    Data packet interval: 125 us
    Bits: 24
```

supporting cases:
* The audioinfo.php didn't show the supported audio format (found by @TheOldPresbyope )
* Camilladsp requires this information to patching configuration files, which failed while the device is playing.

### DIFF
--- a/usr/local/bin/moodeutl
+++ b/usr/local/bin/moodeutl
@@ -274,8 +274,18 @@ function audio_formats($option) {
 			echo $result;
 		}
 	}
-	else {
+	else if($option == '-F') {
 		echo "Audio device busy\n" ;
+		echo shell_exec('cat /proc/asound/card' . $cardnum . '/stream0');
+	}
+	else {
+		$audio_format = trim(shell_exec('cat /proc/asound/card' . $cardnum . '/stream0 | grep Format:'));
+		if (!empty($audio_format)) {
+			echo explode(': ', $audio_format)[1] . "\n";
+		}
+		else {
+			echo "Unable to detect formats\n";
+		}
 	}
 }
 


moodeaudio:
It appears that I2S, HDMI and Headphone devices don't post stream0 information and thus the supported formats can only be obtained when the audio device is closed and alsacap utility is used.

I2S device
```
pi@rp1:~ $ ls /proc/asound/card0/
id  pcm0p
```

HDMI and Headphone
```
pi@rp2:~ $ ls /proc/asound/card0
id  pcm0p
pi@rp2:~ $ ls /proc/asound/card1
id  pcm0p
```

USB device
```
pi@rp2:~ $ ls /proc/asound/card2
id  pcm0p  stream0  usbbus  usbid  usbmixer
```

bitkeeper:
Then I will change it a little bit to:
cat /proc/asound/card1/pcm0p/sub0/hw_params

And include a file exists test:

bitkeeper:
Bump that will also not work because that is the actual settings, not the supported formats.

I will search for something to fix it in the morning.

moodeaudio:
Right. I think what will work is the following:

1. Revert to alsacap method
2. Capture the supported formats during startup when audio device not playing
3. Capture them dynamically but only when changing among HDMI, Headphone or USB in MPD Config

This way the supported formats for the selected audio device will always be available.

bitkeeper:
Instead of trying being to smart, may the easiest thing is in the worker, before the MPD startup, just cache the output of alsacap and use that for static devices like the hdmi, headphone and i2s. And use  /proc/asound/cardx/stream2 for th usb devices ?

moodeaudio:
Even better :-)

I'll make the changes tonight.

PR #352, opened 2021-02-26
[CamillaDSP] cleanup worker job

bitkeeper:
Plus:
- fix wave split with filenames that contains spaces
- fix convolution examples


moodeaudio:
I don't think this is the right fix. The hang issue is caused by some unnecessary session code I added to cdsp.php to get the audio formats. It was closing the session which prevented worker from successfully resetting $_SESSION['w_active'] to 0. This resulted in the waitWiorker() function never returning control to the submitting php script i.e. snd-config.php.

I keep forgetting that worker job loop is like below and wrapped in an open session so job code has access to the session vars. 

```
session_start()

case 'job':
    do something
    break

session_write_close()
```

I've tested what I think is a good fix. I'll push the commit in a bit.

www/command/worker.php
```diff
@@ -1593,25 +1594,16 @@ function runQueuedJob() {
...
 				sysCmd('mpc enable only "' . ALSA_CAMILLADSP . '"');
-
-				sysCmd('systemctl restart mpd');
-				// Wait for mpd to start accepting connections
-				$sock = openMpdSock('localhost', 6600);
-				closeMpdSock($sock);
-				setMpdHttpd();
 			}
 
+			$sock = openMpdSock('localhost', 6600);
```
bitkeeper, line 1604: Without this the 'mpc play' is just a bit to early and fails most of the time.
Prob sleep or every othe cpu cycle waster is good enough.

www/inc/cdsp.php
```diff
@@ -184,13 +181,7 @@ function newConfig($configname) {
...
 
     function detectSupportedSoundFormats() {
-        $session_id = trim(shell_exec("sqlite3 /var/local/www/db/moode-sqlite3.db \"SELECT value FROM cfg_system WHERE param='sessionid'\""));
-        session_id($session_id);
-        session_start();
-        $supported_formats = $_SESSION['audio_formats'];
-        session_write_close();
-
-        $available_alsa_sample_formats_from_sound_card_as_string = $supported_formats; //Sound card sample formats from ALSA
+        $available_alsa_sample_formats_from_sound_card_as_string = $_SESSION['audio_formats']; //Sound card sample formats from ALSA
```
bitkeeper, line 184: @moodeaudio according to your instruction; is this correct ?

moodeaudio:
Looks ok.

Here’s the code exactly as I tested.

    function detectSupportedSoundFormats() {
        $available_alsa_sample_formats_from_sound_card_as_string = $_SESSION['audio_formats']; //Sound card sample formats from ALSA
        $available_alsa_sample_formats_from_sound_card = explode (', ', $available_alsa_sample_formats_from_sound_card_as_string);
        $sound_device_supported_sample_formats = array();
        foreach ($this->alsaToCamillaSampleFormatLut() as $alsa_format => $cdsp_format) {
           if (in_array($alsa_format, $available_alsa_sample_formats_from_sound_card)) {
                $sound_device_supported_sample_formats[] = $cdsp_format;
           }
        }

        return $sound_device_supported_sample_formats;
    }

PR #357, opened 2021-03-01
[CamillaDSP] Add functionality to convert wave files to raw files

bitkeeper:
Use sox for convertion and force 32bit format as output, due an issue with 24bits.

If adds an convert button:
image

- on press an dialog follows to confirm to process or cancel (maybe the texts can be improved)
image
- the wav is convert to 24b raw file.
- if wav is a stereo file it is converted in to mono files.
- - if the conversion is succesfull the orginal wav is deleted.

Also the coeff file info give now a warning if it is wav:
image

See https://github.com/HEnquist/camilladsp/issues/97 about the 24bits issue.

### DIFF
--- a/www/cdsp-config.php
+++ b/www/cdsp-config.php
@@ -161,14 +161,14 @@ else if ($selected_coeff && isset($_POST['info']) && $_POST['info'] == '1') {
 // no implementation required, just a placeholder
 
 }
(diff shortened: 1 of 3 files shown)


bitkeeper:
@moodeaudio 
I could use some help to make the input field with the file name a little bit wider, tried several thing but no succes:
image

moodeaudio:
Another approach is to display the name under the input field. that way if its really long it will just wrap.

PR #391, opened 2021-04-11
Fixes/upnp artwork

bitkeeper:
- Protect upnp_albumart.py url against funky chars.
- Speed up upnp_albumart.py by providing used method 
- Let getUpnpCoverUrl provide mode to upnp_albumart.py

See the three commits notes for more info.

### DIFF
--- a/www/command/upnp_albumart.py
+++ b/www/command/upnp_albumart.py
@@ -11,6 +11,7 @@
 import sys
 import os
 import upnpp
+import urllib.parse
 
 def debug(x):
     print("%s" % x, file = sys.stderr)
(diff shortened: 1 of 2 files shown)


moodeaudio:
Do we still need to append the service type to the name in getUpnpCoverUrl() ?
```
 . $_SESSION['upnp_svctype']
```

bitkeeper:
This PR is independent of that change. I have not paid much attention to upnp_svctype append or not, see the posts of @TheOldPresbyope of the forum.

Only if you don't apeend with getUpnpCoverUrl, you also need to stop appending if to the friendly name. If both don't match you will be in trouble.

PR #392, opened 2021-04-12
Fix broken multiple urls due escaping with upnp_albumart

bitkeeper:
By using escaping for funky urls the multiple urls support is broken. This PR fix that.

### DIFF
--- a/www/command/upnp_albumart.py
+++ b/www/command/upnp_albumart.py
@@ -30,9 +30,14 @@ def artFromMeta(metadata):
     if dirc.m_items.size():
         dirobj = dirc.m_items[0]
         if "upnp:albumArtURI" in dirobj.m_props:
-            o = urllib.parse.urlsplit(dirobj.m_props["upnp:albumArtURI"])
-            url_escaped = urllib.parse.ParseResult(o.scheme, o.netloc, urllib.parse.quote(o.path), "", "", "").geturl()
-            print("%s" % url_escaped)
+            url_orig = dirobj.m_props["upnp:albumArtURI"]
+            urls = url_orig.split(', http')
+            for idx, url in enumerate(urls):
+                if idx > 0:
+                    urls[idx] = "http{}".format(url)
+                o = urllib.parse.urlsplit(urls[idx])
+                url_escaped = urllib.parse.ParseResult(o.scheme, o.netloc, urllib.parse.quote(o.path), "", "", "").geturl()
+                print("%s" % url_escaped)
             sys.exit(0)
 
 


bitkeeper:
Only merge after we get confirmation by @TheOldPresbyope that it fixed the issue with multiple urls.

moodeaudio:
Sure, no prob.

TheOldPresbyope:
Ok, I don't claim exhaustive coverage, but I think I've done enough for you to merge and let the other testers have at it.

So far-

Media Servers: minimserver, Universal Media Server, Synology Media server, also Qobuz (via the control points)

Control Points: BubbleDS (Android), Bubble UPnP (Android), LInn Kazoo (Android), Lumin (iOS), Linn Kazoo (iOS)

OpenHome service enabled.

All combinations appear to work in setting up playlist (e.g., loading the queue), playing tracks, and displaying albumart.

Strangely, I'm having trouble getting albumart displayed using moOde's DLNA server as the Media Server, yet the URL returned by upnp_albumart.py appears to be syntactically correct. I have to wait 'til later to see if it is real.

Similarly, I have to wait 'til later to check UPnP/AV service.

Regards,
Kent

moodeaudio:
Quick test with one pi running dlna server, another upmpdcli set to OpenHome and my iPhone with latest Linn Kazoo and its mostly like below with the "waiting for room..." screen.

IMG_1020

Eventually, after clicking around the UI somehow I get to an album list and can queue something in moOde to play but there is no playlist management. Every track or album clicked is added whether it's already in the Queue or not.

TheOldPresbyope:
I've seen the spinner too but only for a few seconds. Just tried the combination of server and control point you describe. This time I did get albumart but then when I added another album to the queue and started it, the WebUI still displayed the old track info and albumart. I had to refresh manually to bump it. 

As for queue management, all the CPs I've tried seem to add the album playlist as many times as I select it. Lumin at least lets me clear the queue. Candidly, I'm about at my limit of my patience with this OH/UPnP stuff and the varying behavior I see in the various apps.

TheOldPresbyope:
Aha. It turns out I may have been confusing two different actions in the CPs I'm using:
1. From the library listing, add an album or track to the playlist (I keep thinking "queue" ala MPD speak). This will start the first track of the added album (or the single added track) but not always refresh the display if there's already something else playing.
2. From the playlist, select an album or track. This also starts the first track of the selected album or the single selected track and always appears to update the track info and albumart displayed.

bitkeeper:
All other issues aside, I think this PR fix what it should fix. 
So lets merge this one.

moodeaudio:
Will do.

PR #408, opened 2021-06-01
Bump to camilladsp 0.5.2

bitkeeper:
Including:
- camilladsp_plot 0.5.3
- gui 0.4.1
- gui-backend 0.7.1


moodeaudio:
Nice :-)

I don’t have any experience with Camilla but I noticed a pair of level meters and volume control in the Volume section of Pipeline editor. What are they used for?

The volume adjustments didn’t seem to have any effect on sound output.

PR #409, opened 2021-06-04
[mpd] curl input is unstable due used version of libcurl-gnutls

bitkeeper:
Results in unstable radio and upnp playback, which was noticed when starting using alsa_cdsp/camilladsp.

Content:
- This package provides an  alternative version .
- But not for system wide use because that will break other utils.
-That is why it is repacked to a different name and location (/opt/...)
- mpd service user conf script is provided to prefer this alternative version by modifying the LD_LIBRARY_PATH


moodeaudio:
If I understand correctly this provides relief until moOde project migrates to RaspiOS Sid or later which has the upstream fix to curl, right?

PR #424, opened 2021-07-25
Initial implementation of gui for backup restore.

bitkeeper:
@moodeaudio @TheOldPresbyope 

Still missing is:

- ~~The implementation to upload script and set wlanip (is already in the backupmanager and is there in the gui but disabled, until php backend is also implemented)~~.
- Forcing correct file rights after restore if required
- Integration into the system gui of moode, two urls available for backup and restore interface:
  - http://moode?action=backup
  - http://moode?action=restore

http://moode?action=backup preview:

image

http://moode?action=restore preview:
image

With testing the restore mechanisms be aware that you can loose stuff (well you will have it in the backup, but it can't be restored until fixed ;-) if there are still issue with the restore implementation.

### DIFF
--- a/www/backup.php
+++ b/www/backup.php
@@ -30,25 +30,25 @@ playerSession('open', '' ,'');
 if( isset($_POST['backup_create']) && $_POST['backup_create'] == '1' ) {
 	$backupOptions = '';
 	if( isset($_POST['backup_system']) && $_POST['backup_system'] == '1' ) {
-		$backupOptions .= $backupOptions ? ',config' : 'config';
+		$backupOptions .= $backupOptions ? ' config' : 'config';
 	}
 	if( isset($_POST['backup_camilladsp']) && $_POST['backup_camilladsp'] == '1' ) {
-		$backupOptions .= $backupOptions ? ',cdsp' : 'cdsp';
+		$backupOptions .= $backupOptions ? ' cdsp' : 'cdsp';
 	}
 	if( isset($_POST['backup_radiostations_moode']) && $_POST['backup_radiostations_moode'] == '1' ) {
-		$backupOptions .= $backupOptions ? ',r_sys' : 'r_sys';
+		$backupOptions .= $backupOptions ? ' r_sys' : 'r_sys';
 	}
 	if( isset($_POST['backup_radiostations_other']) && $_POST['backup_radiostations_other'] == '1' ) {
-		$backupOptions .= $backupOptions ? ',r_other' : 'r_other';
+		$backupOptions .= $backupOptions ? ' r_other' : 'r_other';
 	}
 
 	if($backupOptions) {
-		$backupOptions = '--what=' . $backupOptions . ' ';
+		$backupOptions = '--what ' . $backupOptions . ' ';
 	}
 
 	$tempBackupFileName = '/tmp/backup.zip';
-	sysCmd('/var/www/backupmananger.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
-	// print('/var/www/backupmananger.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
+	sysCmd('sudo -u pi /var/www/command/backupmanager.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
+	// print('/var/www/command/backupmanager.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
 
 	// create name for backup file in browser
 	$dt = new DateTime('NOW');
--- a/www/backup.php
+++ b/www/backup.php
@@ -50,6 +50,12 @@ if( isset($_POST['backup_create']) && $_POST['backup_create'] == '1' ) {
 		$backupOptions .= '--wlanpwd ' .$_POST['backup_wlan0pwd'] .' ';
 	}
 
+	if( isset($_FILES['backup_scriptfile']) ) {
+		move_uploaded_file($_FILES["backup_scriptfile"]["tmp_name"], '/tmp/script');
+		//TODO: test if file exists
+		$backupOptions .= '--script /tmp/script ';
+	}
+
 	$tempBackupFileName = '/tmp/backup.zip';
 	sysCmd('sudo -u pi /var/www/command/backupmanager.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
 	// print('/var/www/command/backupmanager.py ' . $backupOptions . '--backup ' . $tempBackupFileName);
--- a/www/backup.php
+++ b/www/backup.php
@@ -1,0 +1,160 @@
+<?php
+/**
+ * moOde audio player (C) 2014 Tim Curtis
+ * http://moodeaudio.org
+ *
+ * (C) 2021 @bitlab (@bitkeeper Git)
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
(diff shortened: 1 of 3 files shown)


moodeaudio:
Looks great. Some users also use password to their NAS (Samba).

monkeypox:
Congrats, this is a long standing want for a lot of users.

Is there a use case for someone not wanting to back up or restore the switched options? Seems for the final version you could simplify by just doing everything which would be the right thing almost all of the time.

I haven’t looked to see how this is implemented but it’d be cool if moode could also automatically discover a backup stored on an attached drive and do (or offer to do) the restore on first run.

moodeaudio:
Thats a good point re "just back up everything". One of the challenges with having backup options is that the restore needs to know which options were selected for the backup. For example if backup does not contain moOde radio stations the restore should not allow that option. 

Maybe start out with 
```
WLAN Password
SMB password
Add script 
```

bitkeeper:
@monkeypox yes there are use cases for not wanting to backup or restore everything. If not I would have saved me the time implementing it ;-)

Default a selection is already made; config, cdsp + useer defined radio stations. So in most cased you can just press backup.

But in this ways I can make very selective backups (also concerning size). For example I make backup with only user defined radio stations(mantain those on one single Pi, which I can easly apply to all my other moOde Pi. Also I can share those with other. The output of the backups is fully compatible with the statationmanager (that I have created some moOde version ago).
The Radio station backup/restore at station config can be stripped and is replaced by a single option.
At the same time I can make full backups and still be selective with restore to replace only the part I have broken or want to revert.

I have planned to support automatic restore in the same way as with autocfg; if it fit on the boot partion place it there and rename it to a fixed name.

@moodeaudio you are going to fast haha. This is not for nothing called the intial version hah. Just want publish something underway. The backupmanager.py already has an info command to check the available content of a backup. Only need to time to make use of it to disable/hide certain option.  Make the restore process only somewhat complexer, because we need to upload the backup first and then some the page with the options, so that is why I didn't directly implemented it.


On my list is:
- Finish WLANIP (nothing using WLAN my self, but I know it is important for a lot of others)
- Finish SCRIPT (to apply my customization for my main moode and dev tools for the development station)
- Allow only to restore only options that are available in the backup.
- Allow auto backup restore from backup file at the boot partion.
- Strip radio backup/restore from gui and php (the backupmanager uses stationmanager as derived class, hate redudant code :-)
- Prob fix bugs

SMB pwd will be for an other dev to implement, above is already more that my time allows.

Also there are still a lot of settings not supported by the autocfg system (which is used by the backup/restore), help implementing would be welcome. Especially when creating new functionality we should take care that also the autocfg is implemented, else you you will keep running behind it.

@moodeaudio what is currently the max allowed upload/download size of the webserver? 
Need prob also some checking for that.

moodeaudio:
I don't think there is a max download size but max upload size is defined in two places in php.ini

/etc/php/7.3/fpm/php.ini
```
upload_max_filesize = 75M
post_max_size = 75M
```

I'll try to make some progress on the remaining autocfg settings.

monkeypox:
@bitkeeper

Sounds good, and especially as the camilladsp integrator that’s a valuable option. My pov is just that users are more likely to use software that doesn’t look intimidating and lots of visible options can give that impression so we should try as much as possible to simplify things. Maybe part of the problem is the config css which gives everything equal weight. It might be better if the options were de-emphasized compared to the backup/restore buttons. 

For the wlan, etc. passwords - I know they’re obfuscated in the db, but shouldn’t they still work?

TheOldPresbyope:
I've been keeping quiet 'cuz I haven't had time to think through how I'd use this.

The thing I liked most about the moodecfg.ini approach when it was introduced was I can create a .ini file for my desired configuration from which I can clone as many players as I needed. For each clone, I run a little script to set the specific hostname and possibly different WLAN values and copy the resulting .ini file to the freshly burned moOde "disk". Boot  _et voila._ 

For most of our users, I believe the number one issue is the ability to go from moOde version x to moOde version x+1 and retain all their customized settings in the process.

Note that both of these use cases imply the file being processed on first boot. Like I said, I haven't thought through the implications of a more selective backup/restore that's handled after the player is up and running.

bitkeeper:
Ok less describe what we have for now (sorry for the lengthy read ;-) :

The `backupmanager.py` (www/command) uses of the following for creating the backup:
- stationmanager.py (python import) for the stations
- moodeutl -e /tmp/backup.zip for the system config
- zip write  for the backdrop image
- zip write for /usr/share/camilladsp/config|coeffs 

The `backupmanager.py` (www/command) uses of the following for performing the restore:
- stationmanager.py (python import) for the stations
- autocfg (worke.php) on startup for the system config
- zip reade  for the backdrop image
- zip read for /usr/share/camilladsp/config|coeffs 

Other facts:
- The backupmanager.py, just like with the stationmanager.py, contains a rich interface for manually managing the backup/restore.
- The backup output is 100% compatible with the stationmanager.py.
- A restore without --what will use auto detect for the content to restore.
- When /boot/moodebackup.zip exists on boot it will be automatically restored.
- But it can also be restored from a running system by use of the gui (it will reboot twice in that case)
- when using the backupmanager.py from commandline you need to reboot your self (or not if no system config is restored, but only cdsp and or radio stations)
- And if the autocfg file is also present followed by the autocfg. An autocfg file will in this way always overrule the settings in teh backup.
- A custom script can be added to the config( but the use is implemented by the autocfg system, see #418)
- autocfg DOESN'T support backup of the wlanpwd, but a plain password can be added to ini file by hand.
- autocfg DOES backup the passwords of the sources. 

**Proposed changes/actions:**

- I will remove any custom backup/restore options from the gui, including wlanpwd and script. Use the commandline version if you require a custom backup/restore, scripts etc.
- I deliver the functionaly as two urls:
  - http://moode/backup.php?action=backup
  - http://moode/backup.php?action=restore
 - Tim can integrate two buttons from the system config page to calls this urls or integrate the functionality entirely in a location of his choice, that is up to Tim :-)

Is that a go?


<br/><br/>

For completeness all available options of the backupmanager for now:
```
pi@moodep4:/tmp $ /var/www/command/backupmanager.py --help
usage: backupmanager.py [-h] [--version]
                        [--what {config,cdsp,r_sys,r_other} [{config,cdsp,r_sys,r_other} ...]]
                        (--backup | --restore | --info) [--db DB]
                        [--script SCRIPT] [--wlanpwd WLANPWD]
                        backupfile

Manages backup and restore of moOde configuration.

positional arguments:
  backupfile            Filename of the moode backup.

optional arguments:
  -h, --help            show this help message and exit
  --version             show program's version number and exit
  --what {config,cdsp,r_sys,r_other} [{config,cdsp,r_sys,r_other} ...]
                        Indicate what to backup/restore. (default for backup:
                        config cdsp r_other, default on restore: auto detect
                        content)
  --backup              Create backup.
  --restore             Restore backup.
  --info                show information what is included in the backup
                        (detect available 'what')
  --db DB               File name of the SQL database. (default:
                        /var/local/www/db/moode-sqlite3.db
  --script SCRIPT       Add script file to the backup (is executed when config
                        is restored)
  --wlanpwd WLANPWD     When creating a backup, supply a password for wifi
                        access (applied when restoring the backup)

Root privileges required for restore.
```

moodeaudio:
Lets go for it :-)

moodeaudio:
After working with this for a while I'm going to go with the original version that provides the options in the UI. There is a small bit of cleanup to do, a few enhancements and one part that I'm having difficulty with. 

I'll post a status update in a few days.

PR #440, opened 2021-09-19
[camilladsp] Quick select of camilladsp configuration in top header

bitkeeper:
**Case:**
On my main listing system I use multiple headphones and speakers.
Each uses one or more camilladsp configs. Before start listing I want to know which config is active and if needed swich to/between different configs.

**Usage:**

When camilladsp is turned off everything is still the same.

But when CamillaDSP is turned on the name of the configuration appear on top left fromthe main menu:

image

When you click the config a popup menu appears where you can select a different config (current active is indicated with a speaker icon):

image

When switching between configs the change is almost right away.

### DIFF
--- a/www/command/moode.php
+++ b/www/command/moode.php
@@ -822,6 +822,29 @@ else {
 		case 'clientip':
 			echo json_encode($_SERVER['REMOTE_ADDR']);
 			break;
+
+		case 'camilladsp_setconfig':
+			if ( isset($_POST['cdspconfig']) ) {
+				require_once dirname(__FILE__) . '/../inc/cdsp.php';
+				$cdsp = new CamillaDsp($_SESSION['camilladsp'], $_SESSION['cardnum'], $_SESSION['camilladsp_quickconv']);
+				$currentMode = $_SESSION['camilladsp'];
+				$newMode = $_POST['cdspconfig'];
+				playerSession('write', 'camilladsp', $newMode);
+				$cdsp->selectConfig($newMode);
+				if ($_SESSION['cdsp_fix_playback'] == 'Yes' ) {
+					$cdsp->setPlaybackDevice($_SESSION['cardnum'], $_SESSION['alsa_output_mode']);
+				}
+
+				if ( $_SESSION['camilladsp'] != $currentMode && ( $_SESSION['camilladsp'] == 'off' || $currentMode == 'off')) {
+					submitJob('camilladsp', $newMode, 'CamillaDSP ' . $cdsp->getConfigLabel($newMode), '');
+				} else {
+					$cdsp->reloadConfig();
+				}
+			}
+			else {
+				$msg = 'Error: missing camilladsp config name';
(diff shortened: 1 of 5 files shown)


moodeaudio:
Quick select is a great idea.

What I would suggest for the UI is the following:

- Place a text icon "dsp"in the center of the header that activates the selection menu. This will allow the feature to fit on mobile layouts. There is an example of center icon placement in header.php, "<div id="multiroom-sender"
- It would be preferable to use a checkmark to the right of the item instead of a speaker icon on the left. This is so its consistent with selection on the context menus.
-

bitkeeper:
Selected config Icon is replaced by standard check of fa.
image


Concerns the drop down menu label (which is the name of the current cdsp config) I made the following changes:
- Added a chevron-down icon to inidacte that it is a drop down menu.
- Used the same font style as with the Library menu (size depends on display)
- Prefer the drop down menu on the right side of the screen so kept that location. To present when in the middle.
- Checked with some mobile in portrait and landscape mode to check if it fits.

image

So I would say give it a spin and if you don't like we can always change or revert it :-)

Found one issue; with a lot of cdsp configs; the popup menu doesn't scroll. So if you how to enable scrolling please let me know.

bitkeeper:
Forgot to say that the popmenu is only available on the playback page. Prevents overlap with text of the library filters.

moodeaudio:
Sounds good. I'll merge later today.

Re popup menus not scrolling. Is this happening on the CamillaDSP Config screen?

moodeaudio:
Odd. I merged this using Github Desktop which just works but the PR still shows as open in the repo.

bitkeeper:
I see that the merge isn't successfull, because only 1 out of 3 commits is merged to the develop branch. Because after the merge the code it directly altered it will be very hard to merge the remaining two commits.
Prob I will have to intergrate and merge this by hand in a speperate PR.

moodeaudio:
Prolly not needed based on the most recent commits for the Quick Select feature.

bitkeeper:
The most recent commits broken the correct behaviour; switching from configurations while playing the song.
Also the name of the current config with chevron-down icon is lost.
And turning cdsp off from the quick selected hides the icon (makes ab test hard), which was indented only when it was intial turned of.

moodeaudio:
Breakage: switching from configurations while playing the song.
- Ok I see what you were going for. The breakage is in moode.php and is caused by the job being submitted for filter changes (this restarts MPD). I'll add back the "off" logic and we should be ok.

Also the name of the current config with chevron-down icon is lost.
- Long names won't fit on small mobiles like iPhone SE. The only viable option is to have an icon.

And turning cdsp off from the quick selected hides the icon (makes ab test hard), which was indented only when it was intial turned of.
- I'll fix.

bitkeeper:
No need any more; fixed and merged already ;-)

Thanks for the other cleanup stuff you did to the PR! :-)

PR #442, opened 2021-09-30
Fix for "add favorite" duplicated + some missing items autoConfig

TookaFace:
Fix for "add favorite" duplicated with the heart icon
Add some missing items in autoConfig


moodeaudio:
Better to split into two separate PR's.

I think the Add fav issue is more involved and has some JS logic associated with it. I'll investigate.

TookaFace:
Ok, i close this, i will open separate PR

PR #448, opened 2021-10-09
Added powersave governor.

tibrezus:
I have my moode installed on a rpi3 which is connected to a 5V/0.5A power source. I use it because is very low noise power source and is also very convenient. It works fine 99% of the time while listening high-res files, but when I try to update volumio and the CPU usage goes high files on the SD-card get corrupted because of insufficient power. 
Limiting the multiplier to it's minimum solves this issue for me, but is very annoying to configure it manually and setup a crontab to overwrite the moode configuration. So I added the powersave config on the System Config to have a fast solution to whoever incurs in SD card data corruption due to low current power source.
It also heats less and 600Mhz are more than enough to handle all the workloads in moode.
I don't have the time to test it, please give it a try.
Cheers

### DIFF
--- a/www/sys-config.php
+++ b/www/sys-config.php
@@ -329,6 +329,7 @@ $_select['browsertitle'] = $_SESSION['browsertitle'];
 // SYSTEM MODIFICATIONS
 
 // CPU governor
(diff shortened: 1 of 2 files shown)


moodeaudio:
Sounds like your issues are very unique to your particular configuration, especially sd card corruption which is rare for myself and the vast majority of moOde users.

PR #463, opened 2021-11-27
Added currentsong.json.php

puzon:
Script is returning /var/local/www/currentsong.txt as json object.

It's helpful with integrating moode with e.g. Home Assistant to detect playing state (mpd, spotify, etc.)


moodeaudio:
What is "Home Assistant" ?

puzon:
[Home Assistant ](https://www.home-assistant.io/) is open source home automation system.

I'm using script from this PR in automation that turns my amp on or off depending on moode playback state.

I previously created script directly on player and lost it after a system update, so I thought it would be a good idea to add this to the repository. Maybe this will also help others who use Home Assistant or similar software.

moodeaudio:
If you put your script in /var/local/www/commandw it won't be overwritten by in-place updates. Scripts in www/command are to support generally applicable features or system functions.

I'm curious why json format is required by Home Assistant software. Can the curentsong file just be parsed as-is?

puzon:
Oh, i didn't know about /var/local/www/commandw.

Home Assistant can fetch string value (as is) or json object using Jinja2 templating engine. Extracting value from json to sensor is as simple as writing `{{ value_json.file }}` ([ref](https://www.home-assistant.io/integrations/sensor.rest#external-ip-address)).
I think that it can't easily parse currentsong.txt with current format.

So i thought json format will be most universal. And i think this data exposed by moode without requiring to manually upload any additional script may be a good thing. Especially because moode already exposes currentsong command from mpd service :)

moodeaudio:
I think the one liner below might do it. I'll add to the TODO list to investigate.

```
json_encode(explode("\n",file_get_contents('/var/local/www/currentsong.txt')));
```

puzon:
Your one liner will return a json array with lines from file instead of key:value object.

Shorter than original from PR and still works:
```$file = '/var/local/www/currentsong.txt';
if (file_exists($file)) {
    try {
        preg_match_all('/^(.+?)= ?(.*?)$/m', file_get_contents($file), $matches);
        print(json_encode(array_combine(array_map('trim', $matches[1]), $matches[2])));
    } catch (\Exception $exception) {
    }
}
```

Input:
```
file=https://n04a-eu.rcs.revma.com/1nnezw8qz7zuv?rj-ttl=5&rj-tok=AAABdZRNZYAALJ0Ql0pJg9g41Q
artist=Radio station
album=MUZO.FM
title=ISAIAH COLLIER & THE CHOSEN FEW - PART I FORGIVNESS
coverurl=imagesw/radio-logos/MUZO.FM.jpg
track=
date=
composer=
encoded=VBR
bitrate=128 kbps
outrate=16 bit, 44.1 kHz, Stereo, 1.411 Mbps
volume=12
mute=0
state=play
```

Output:
```
{
	"file": "https:\/\/n04a-eu.rcs.revma.com\/1nnezw8qz7zuv?rj-ttl=5&rj-tok=AAABdZRNZYAALJ0Ql0pJg9g41Q",
	"artist": "Radio station",
	"album": "MUZO.FM",
	"title": "ISAIAH COLLIER & THE CHOSEN FEW - PART I I I UNDERTANDING",
	"coverurl": "imagesw\/radio-logos\/MUZO.FM.jpg",
	"track": "",
	"date": "",
	"composer": "",
	"encoded": "VBR",
	"bitrate": "128 kbps",
	"outrate": "16 bit, 44.1 kHz, Stereo, 1.411 Mbps",
	"volume": "12",
	"mute": "0",
	"state": "play"
}
```

PR #469, opened 2021-12-15
Enable optional Spotify/librespot login

GeoffMaciolek:
## Overview

Add config options in Spotify to allow login

## Details
This configuration change *should* be all that's needed to enable [librespot](https://github.com/librespot-org/librespot) (the Spotify Connect client) to support user logins. This helps in situations where the local discovery service isn't working, and also (I believe) allows for "remote access" playback. (Tested here with a device accidentally on the wrong wifi network!)

## WARNING

It's a huge faux-pas, but I have **not** tested this! I repeat: **I have not tested this code!** It's borderline-trivial, but, I wanted to put in this PR before investing time in the testing process, in case the PR in general isn't considered worthwhile, adequate, etc.  (I have tested the generated command-line itself, and, the credential caching works well - and your code already handles optionally clearing it!)

### DIFF
--- a/boot/moodecfg.ini.default
+++ b/boot/moodecfg.ini.default
@@ -187,6 +187,8 @@ spotify_normalization_knee = "1"
 spotify_format = "S16"
 spotify_dither = ""
 spotify_volume_range = "60"
+spotify_username = ""
+spotify_password = ""
 
 [Squeezelite]
 squeezelite_PLAYERNAME = "Moode"
--- a/var/local/www/db/moode-sqlite3.db.sql
+++ b/var/local/www/db/moode-sqlite3.db.sql
@@ -477,6 +477,8 @@ INSERT INTO cfg_spotify (id, param, value) VALUES (12, 'normalization_knee', '1'
 INSERT INTO cfg_spotify (id, param, value) VALUES (13, 'format', 'S16');
 INSERT INTO cfg_spotify (id, param, value) VALUES (14, 'dither', '');
 INSERT INTO cfg_spotify (id, param, value) VALUES (15, 'volume_range', '60');
+INSERT INTO cfg_spotify (id, param, value) VALUES (16, 'username', '');
+INSERT INTO cfg_spotify (id, param, value) VALUES (17, 'password', '');
 
 -- Table: cfg_network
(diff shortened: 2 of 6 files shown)


GeoffMaciolek:
Hey there - any comments on why this was closed? (Don't worry - I don't feel entitled to have my patch accepted, but, I was hoping at least for a quick look at my code, or a comment on if this feature would be useful or not.)

I will likely attempt more testing, and will attach screenshots, but, I hope to hear back.

**Note:** I just saw that "Allow edits by maintainers" was not checked - I think the default for that is set up on your repository, but, if that was the cause of the "closing" then that's resolved.

moodeaudio:
It was closed because you stated that you didn't even bother to test the code.

Besides that problem the usage scenario "...situations where the local discovery service isn't working..." is rare based on the lack of trouble reports on our support Forum concerning this type of issue and thus I don't see a need to add anything.

GeoffMaciolek:
Tim - I'm happy to run tests soon, but I'm reluctant to put too much work
into it, given you don't see it as a useful feature. I could have started
this as an issue report, but I thought it would be better received as an
actual PR.

As for the reports? I suspect some people did what I did - SSHed in and
performed a login, and let the cached credentials work from there.

The feature useful to me, as remote control is useful, especially in places
with mediocre or intermittent connectivity (where people may be more likely
to use hotspots or turn their phone's wifi off - the situation I was in.
Intermittent/slow connectivity is generally adequate for a 320kbps stream,
but can be painful enough that folks will switch to mobile data.)

Anyway, apologies if submitting this untested was frustrating or insulting;
I'll test this soon. Would you like me to submit a new PR, or request this
one be reopened?

monkeypox:
Geo,

No project takes code that hasn’t been tested so that’s your first step and that will let you see if their are any issues and get them sorted before asking someone else to take the time to review your code.

It would probably be helpful if you registered an account with the forum to see if other members have had similar problems to demonstrate that this addresses a more general problem that has simply flown under the radar.

chaseadam:
Found this while searching for "login" support on librespot. This is handy when there is a "player" device which solely uses the spotify API to control a device (vs local discovery) or is on a network which is different than the player.

PR #476, opened 2022-01-05
Small fixes for bullseye

bitkeeper:
Small fixes for bullseye:
    * No check of mpd patch_id anymore
    * Check version of python RPI GPIO with dpkg instead


moodeaudio:
Good stuff.

PR #480, opened 2022-01-18
Cleanup for bullseye

bitkeeper:
Collecting files that can be remove from the moOde repo for the upcoming bullseye build.

Removing:
* Thirdparty software and patched that are not needed anymore or are move to pkgbuild
* Configuration files not needed anymore


moodeaudio:
The "Cleanup not needed configurations files" are the ones where we just use the pkg default correct?

bitkeeper:
Yep that is correct.

The remaining configuration files can be grouped as:

1.  Just copy (ownership of moode audio player)
2. Don't copy, in the repo just an example file,  edit the actual config with sed
3. Overwrite existing file suplied by other package (no ownership of moode audio player)

Too make it the packing scripts easier and prevent handpicking of all the files for package creating, I'm considering adding as postfix to category 2&3. For example `/etc/foo.conf.sed` and `/etc/foo.conf.overwrite`.

What do you think about it?

moodeaudio:
If it makes things more easily understood then good idea.

bitkeeper:
@moodeaudio The main cleanup is finished. If you are ready for it .. you can "squash and merge" it.
The pkgbuild moode-package build scripts is based on this changes.

PR #482, opened 2022-01-23
Always generate mpd.conf on startup of worker

bitkeeper:
This makes it possible to:
* Not need any long to  provide a default mpd.conf
* Also always have an up2date config (after a backup restore)
* Restore the functionality of Developer tweaks(mpd config merge) again

But also skip device handling when USB is used to not break the USB device handling.

### DIFF
--- a/www/command/worker.php
+++ b/www/command/worker.php
@@ -335,21 +335,28 @@ else {
 workerLog('worker: -- Audio config');
 //
 
+$ignoreUpdateMpdDevice = False; // when true ignore update device, only generate mpd conf
 // Update MPD config
 if ($_SESSION['multiroom_tx'] == 'Off') {
 	// NOTE: Only do this for I2S (non-hotplug devices) or if in-place update applied
 	if ($_SESSION['i2sdevice'] != 'None' || $_SESSION['i2soverlay'] != 'None' || $_SESSION['inplace_upd_applied'] == '1') {
-		updMpdConf($_SESSION['i2sdevice']);
+		$ignoreUpdateMpdDevice = False;
 		$mpd_conf_upd_msg = 'MPD conf updated';
 		playerSession('write', 'inplace_upd_applied', '0');
 	}
 	else {
-		$mpd_conf_upd_msg = 'MPD conf update skipped (USB device)';
+		$ignoreUpdateMpdDevice = True;
+		updMpdConf($_SESSION['i2sdevice'], True);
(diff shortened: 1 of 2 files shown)


moodeaudio:
After testing there are some logic issues and other breakage so better to revert the commit for now and look at how to accomplish always updating mpd.conf during startup while at same time avoiding device change when USB device is configured.

bitkeeper:
Can you be a little more specific about the issues? It is hard to search for 'some logic issues an other breakage' ;-)
For i2s device there should be no behaviour changes at all (it did already write the mpd conf always).

Instead of reverting the change I would prefer fixing it.

Can you see if you change the line 1983 in playerlib.inc of the problem is solved ?:
from:
```php
if ( $adevname != '' || $ignoreNotFoundDevice == False ) {
```
to:
```php
if ( $ignoreNotFoundDevice == False ) {
```

With that change the default behaviour of `mpdUpdConf` is exactly the same as it before the patch was.
Only when a second argument with the value `True` is provided it will behave different.

The leaves only the work.php where a `True` is provided to behave different. All uses in other locations in moode will not be affected.

moodeaudio:
My apologies, it was late last night and I didn't feel like writing a lot but here are the issues.

worker.php
```
- bad var name on line 357: $updateMpdDevice should be $ignoreUpdateMpdDevice
- duplicate execution of updMpdConf(): lines 349 and 357
```
it wouldn't have been obvious but always running updMpdConf() during startup breaks a couple of things
```
- Multiroom Sender ALSA config
When sender is turned on it reconfigures _audioout.conf and _sndaloop.conf to hw:3,0 (3 = PCM Dummy device)
If updMpdConf() gets run during startup the files will get reverted to "plughw:0,0" and this sometimes causes MPD to error out with "fail to open ALSA device..."

- Stored mixer name for configured but unplugged or turned off USB DAC
$_SESSION['amixname'] is assigned the string "Invalid card number."
This requires user to reboot after plugging in or turning on DAC
```

The cfg_system param inplace_upd_applied = 1 will cause updMpdConf() to be run during startup and so maybe this could be used to satisfy the three goals of the PR.

Set it as part of apt installing moode-player
Set it as part of Restore
Set it to have dev tweaks applied

moodeaudio:
I pushed a commit to worker.php earlier that improves the readability of the code block that handles updating mpd.conf during startup. It should make to easier to modify and test if needed. https://github.com/moode-player/moode/commit/80e49219773ba603b3c1093a035e0dd5d921e6a5

PR #483, opened 2022-01-24
Improve mpd update conf

bitkeeper:
* Instead of the hard to read nested ifs use a switch/case statement.
* Handle device type in separate function  `playbackDestinationType`.
* Generate mpd.conf also when file is missing.


moodeaudio:
Looks good. I'll pull down in a bit and do some testing :-)

PR #495, opened 2022-02-24
Add IPv6 listen directive to the nginx config

kburdett:
The current configuration does not bind to port 80 for IPv6 interfaces. Adding a new listen directive for IPv6 will allow this server to better function in a dual-stack environment.


moodeaudio:
NGINX IPv6 listener is part of release 8.2.5

PR #536, opened 2022-08-07
feat: add inotify to minidlna

vrenaville:
* Purpose
With local drive, using the inotify option, permit to not rebuild the database
when we add new songs/playlist on disk

* If needed I can try to add a button to toggle on/off this functionnality

* I flag it

### DIFF
--- a/etc/minidlna.sed.conf
+++ b/etc/minidlna.sed.conf
@@ -92,7 +92,7 @@ friendly_name=Moode DLNA
 #model_number=
 
 # Automatic discovery of new files in the media_dir directory.
-inotify=no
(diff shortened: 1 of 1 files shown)


moodeaudio:
Yes, it will need an ON/OFF toggle button because some users may not want it ON.

Probably just parse out the value of the option from the conf file when loading the PHP template, then to update it use sed -I in the worker.php job 'minidlna'

To implement this requires updating several scripts. I can provide some guidance and sample code just let me know :-)

moodeaudio:
Closed due to lack of activity. Will reopen of needed.

PR #539, opened 2022-08-27
Fix flac coverart may not be displayed

yn386:
This PR fixes following problem.

### Problem:
Flac coverart may not be displayed.

### moOde Version
Release 8.1.2

### Environment
Raspi: Pi-3B 1.2 1GB
Audio: Pi HDMI 1

### How to reproduce:
Play flac file that have 7 or more metadata block before PICTURE metablock.

### Detail of changes:
To prevent corrupt header from causing endless loop, only first 7 metadata blocks are processed and others are ignored here:
https://github.com/moode-player/moode/blob/0fbbe1e76cbfcfa2a2c1d44ecb99398b0123de18/www/inc/Zend/Media/Flac.php#L171-L175
but sometimes flac has multiple APPLICATION metadata block like this
```
[STREAMINFO]
[VORBIS_COMMENT]
[SEEKTABLE]
[APPLICATION]
[APPLICATION]
[APPLICATION]
[APPLICATION]
[APPLICATION]
[PICTURE]
[PADDING]
```
This is not corrupted flac and compatible with flac specification.

So I made changes instead of counting metadata block to prevent endless loop with coruupted files.

- Check whether metadata indicates 'invalid'.
According to the flac specification, BLOCK_TYPE 127 indicates 'invalid' to avoid confution with a frame sync code.
So if metadata block type was 127, the flac is able to be regarded as corrupted file.

- Check whether current file offset is in the file.
If Last-metadata-block flag is still 0 until the end of file, the flac is able to be regarded as corrupted file.

### Effect:
Coverart is displayed when playing flac file that have 7 or more metadata block before PICTURE metablock.

### DIFF
--- a/www/inc/Zend/Media/Flac.php
+++ b/www/inc/Zend/Media/Flac.php
@@ -67,6 +67,9 @@ final class Zend_Media_Flac
     /** The picture metadata block */
     const PICTURE        = 6;
 
+    /** The invalid metadata block */
+    const INVALID	 = 127;
+
     /** @var Zend_Io_Reader */
     private $_reader;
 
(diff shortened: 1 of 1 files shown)


moodeaudio:
Looks like a nice fix :-)

I'll merge and test this weekend.

PR #542, opened 2022-09-06
Fix librespot volume_range max value.

tweisberger:
Making fix to match librespot's [volume_range max value](https://github.com/librespot-org/librespot/blob/dev/src/main.rs#L325). If entering value more than 100 in moOde's settings, librespot will fail to open.


moodeaudio:
Thanks :-) I'll merge tonight.

PR #543, opened 2022-09-10
Fix artist may be displayed as undefined or null

yn386:
This PR fixes following issue.

### Issue:
When playing audio file that does not have id3tag,

- Playlist shows 'undefined' as an artist.
- Player shows 'null' as an artist.

### moOde Version
Release 8.1.2

### Environment
Raspi: Pi-3B 1.2 1GB
Audio: Pi HDMI 1

### How to reproduce:
Play mp3 or flac file that does not have id3tag.

### Detail of changes:
Playlist shows artist at
https://github.com/moode-player/moode/blob/69eabe8d63f9f7ab285bb7ad35936ea7c2fd6284/www/js/playerlib.js#L1254
but AlbumArtist may be undefined.
To prevent showing 'undefined', I added checking whether AlbumArtist is undefined or not.

Player shows artist at
https://github.com/moode-player/moode/blob/69eabe8d63f9f7ab285bb7ad35936ea7c2fd6284/www/js/playerlib.js#L872-L873
but albumartist may be null.
To prevent showing 'null', I added albumartist element to the array that is returned from enhanceMetadata().

### Effect:
Show artist as 'Unknown artist' even if playing file that does not have id3tag.

### DIFF
--- a/www/inc/mpd.php
+++ b/www/inc/mpd.php
@@ -663,6 +663,7 @@ function enhanceMetadata($current, $sock, $caller = '') {
 		} else {
 			// Song file, UPnP URL or Podcast
 			$current['artist'] = isset($song['Artist']) ? $song['Artist'] : 'Unknown artist';
+			$current['albumartist'] = isset($song['AlbumArtist']) ? $song['AlbumArtist'] : 'Unknown artist';
 			$current['title'] = isset($song['Title']) ? $song['Title'] : pathinfo(basename($song['file']), PATHINFO_FILENAME);
 			$current['album'] = isset($song['Album']) ? htmlspecialchars($song['Album']) : 'Unknown album';
 			$current['disc'] = isset($song['Disc']) ? $song['Disc'] : 'Disc tag missing';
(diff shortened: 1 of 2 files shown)


moodeaudio:
Files without any tags not only break the display of metadata in the Queue they also break the display of items in Library views. Given that the majority of our user base has tagged collections I don't feel a pressing need to add code for these types of files.

Why not just tag the files?

yn386:
Hi, @moodeaudio. Thank you for your reply.

> Why not just tag the files?

Indeed tagging files is correct to manage libraries...
I often download media files distributing as free, and sometimes those don't have tag.
To keep original, I don't add tag or modify such files.

Album name is displayed as 'Unknown album' even if the playing media does not have id3tag, so I thought displaying 'Unknown artist' is more better than displaying 'undefined' or 'null'.

moodeaudio:
I suppose if the Library breakage could be addressed I'd be willing to reconsider adding code to handle untagged files.

PR #544, opened 2022-09-21
[backup_manager] Fix error restore when restoring backup without stations

bitkeeper:
The restore always checked if the backup did contain valid stations, which isn't the case with this isue.
Fix is just skip chekc in this case.


moodeaudio:
Worked perfectly. User reported success :-)

PR #552, opened 2022-10-09
Various css fixes

Alberto-Bulpros:
Fixed long artist name overflowing from album-art in tag-view, if selected and has more than one album


Alberto-Bulpros:
Pressed enter by mistake... also centered the "Cancel" button by 0.5em (according to ita margin/padding) to the left in the big-volume view.

moodeaudio:
I'm not able to repro either of these issues. 

Post some screen shots from your end. It would help me understand what might be going on.

Alberto-Bulpros:
Hi Tim,

it's a little difficult... now that I have the changes in place; but I'll try nevertheless.

Anyway:


  1.  the long artist name... I have 5 albums from the band BADBADNOTGOOD, and if I just select the artist, in the TAG view, I get the album-art of the first album, with a 50% opacity overlay showing the artist's name. As it is a long single word, it does not wrap on multiple lines within the frame of the album-art, instead it gets written all-in-one-line, which results in it overflowing on the right-hand side, towards the track-list.
It depends on the dimensions of the screen (and therefore of the album-art box too), though; as I can see it with the localUI 7" display, or in my browser at 50% screen width window, but not in my browser full-screen.
  2.  The "Cancel" button in the volume adjustment screen (accessible either from the TAG view, or the screensaver view) is not horizontally centered: to see it, set the volume to 100%, and sight the gap of the canvas between the two ends of the "volume level circle", and the word "Cancel" in the button; you'll notice it is slightly moved to the right.
This is because it has a 1em margin, but its left property is translated to the left by calc(50% + 0.5em), instead of the correct calc(50% + 1.0em)

You should be able to reproduce them both, especially the cancel button one; it may be more evident with landscape orientation, though, like a tablet, instead of a mobile phone in the - more - usual portrait mode.

P.S.
There are also other small glitches which I have fixed (but not committed / pull-request'ed yet), like:
1. If there is nothing playing, the "-" between Artist and Title is still visible. (JS side)
2. If there is nothing playing, and you display audio information, in the Playback tab the "Encoded at" item shows "File does not exist", instead of the correct (as the Decoded to and Output rate ones) "Not playing" (PHP side).
3. If you are playing something, and from the screensaver view you clear the queue, the progress bar does not reset to 0, nor it does hide (it is not shown if you startup not playing, but it remains there if you play, then clear the queue...) (JS side)
4. The same progress bar does not reset properly if while playing (again, in the SS view) you press the "|<" (back) button, although the audio properly stops, and the counter resets to 00:00 (this one I still need to look into...) (JS side, I guess)
5. The album art and the overlaying button, in the tag-view, have different sizes, and the album-art especially is too tall, and unpleasantly overflows onto the underlying text about number of albums/duration/etc. (again, only in the case where the artist has several albums - see point 1. about word-wrapping) (CSS side)


Cheers, Al.

________________________________

Alberto-Bulpros:
Here is my best with the pictures...

Cheers, Al.
________________________________

Alberto-Bulpros:
P.S.

I need to think something, as my Git account was set up from within the company I work for, and I receive all e-mails there...
Maybe the best is to setup a new account, personal this time, and redo everything from there... I'll let you know, or... well, you'll probably see yourself.


Cheers, Al.
________________________________

moodeaudio:
I didn't see any screen shots but based on your description I am able to repro both issues and your proposed CSS fixes work just fine.

I have another commit I'm working on today and after thats pushed I'll merge your PR :-)

Alberto-Bulpros:
I did attach a zip file in a later e-mail, anyway, since you have seen the issues, that's ok.

Also, as I wrote, I have found (and fixed) other small issues, which you can reproduce as well; we can look into these and eventually discuss them, once we both have time; meanwhile they are committed in my system.


Cheers, Al.
________________________________

moodeaudio:
You need to submit individual PR's for each proposed change. Mixing fixes and new features and adding out of scope changes to an existing PR is not acceptable.

For example  this PR titled "Various CSS fixes" has now encompassed PHP and JS files, comment tags and adding extra metadata to CoverView.

Also with each PR that involves a new feature or enhancement there needs to be a description that explains the justification for the change. For example why should full extra metadata be added to CoverView?

Alberto-Bulpros:
Hi Tim,

I just merged in the wrong direction, and everything went into the branch that had the pull-request.

Ignore that, later I'll delete everything, create a new account, make a new clone and start the changes there.


Cheers, Al.
________________________________

PR #554, opened 2022-10-11
Add metadata comment

AlbertoNarduzzi:
Added comment to the enhanceMetadata function in order to be able to retrieve it in the now-playing context.
Also, added its visualization in the TAG and ALBUM views (only if the preferences option is enabled) after the album's (YEAR), and styled accordingly.

### DIFF
--- a/www/css/panels.css
+++ b/www/css/panels.css
@@ -224,8 +224,8 @@ img.coverart {border-style:none;border-radius:.1em;box-shadow:0 .1em .2em rgba(3
 #top-columns .album-name-art, #top-columns .album-name {display:block !important;vertical-align:middle;min-width:25vw;}
 #top-columns .album-name-art {line-height:1.25em;}
 #albumsList .lib-entry span.artist-name {display:none}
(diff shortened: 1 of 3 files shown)


AlbertoNarduzzi:
I use it mainly for the type/release of the album, e.g.
Live / Compilation / EP etc. or to differentiate remasters / releases for
different Countries, such as US 1986 release / UK 1985 release / 2001
remaster etc.

moodeaudio:
Thats somewhat close to the original intent of the "Include comment tag" Prefs setting. It's not obvious in the help for the setting but it was meant to be used in conjunction with Prefs "Album key" set to "Album@Artist@AlbumID". 

Are you also setting Album key to that value?

AlbertoNarduzzi:
> Are you also setting Album key to that value?

no, I'm not, it's enough for me to be able to identify them this way, in either the tag, or the album view (most of the times they do have the very same album-art, so no other quick way...).
I cannot say if the comment then takes place in the sorting key (you might say a word, though), as I do have some albums that are released in different Countries in the same year, quite a few months apart, and may get sorted "wrongly"... I haven't investigated yet if I indeed might face such an issue.

PR #555, opened 2022-10-11
Pause rewind stop playbar fixes

AlbertoNarduzzi:
Clearing the play queue changes the player status to STOP, which shouldn't show the play-bar/progress-bar (as may be seen on startup, if nothing is in the queue).
Also, pressing the BACK/PREV button, especially on the screen-saver view, didn't reposition the tick on the progress-bar to its "zero" position: pressing PREV during playback => rewinds to zero and puts the player into PAUSE.


moodeaudio:
This PR contains out of scope commits from PR #554 

Resubmit so that it only covers `Pause rewind stop playbar fixes`

PR #556, opened 2022-10-11
Pause rewind stop playbar fixes

AlbertoNarduzzi:
Removed the dash between artist and track if empty (nothing playing)
Hidden the progress-bar upon STOP (clear queue)
Properly repositioned the progress-bar upon pressing BACK/PREV while playing: resets song to zero time and enters PAUSE mode.

### DIFF
--- a/www/inc/mpd.php
+++ b/www/inc/mpd.php
@@ -586,6 +586,7 @@ function enhanceMetadata($current, $sock, $caller = '') {
 	$current['performer'] = $song['Performer'];
 	$current['albumartist'] = $song['AlbumArtist'];
 	$current['artist_count'] = $song['artist_count'];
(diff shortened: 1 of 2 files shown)


moodeaudio:
This still has commits from the other PR...

AlbertoNarduzzi:
Hi Tim,

I don't know how this could be possible, as I create a new branch (based on
develop, from moode-player/moode), add the changes to it, then create a
pull request.

Then again, another new branch from the same develop, so it's clean, and
proceed with the 2nd PR...

I believe this to be correct, as if I check-out any of such branches, I end
up with ONLY the changes involved in them. I use VisualStudioCode to handle
the repo, though... but from my perspective it does its job.

If you have a better strategy, I'll be happy to follow it.

Cheers, Al.

PR #558, opened 2022-10-11
OSK for 7-inch display

AlbertoNarduzzi:
Added an on-screen-keyboard suitable for clients without an operating-system one, such as the official Raspberry Pi 7" touch screen. It self-installs ONLY if detects Chrome for ARM platform, so browsers / phones / tablets should not be affected in any way.

### DIFF
--- a/www/css/osk.css
+++ b/www/css/osk.css
@@ -1,0 +1,111 @@
+/*
+ * On-screen keyboard. (C) 2022 @Nutul (albertonarduzzi@gmail.com)
+ *
+ * This Program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * This Program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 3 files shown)


moodeaudio:
Rather than do an auto install it might be better to have a button n the Local display section of System config to Install / Remove the OSK. Many users that have a locally attached display only use it for viewing the UI and not interacting with it. These displays don't have touch capability.

I can add the option if it's decided to go that way.

AlbertoNarduzzi:
Hi Tim,

I for one even prefer this option to the autorun one. If it is for me,
let's go this way.


Cheers, Al.

moodeaudio:
Hi Al,

The latest commits have the OSK working on my end. It's super nice :-)

One thing I would suggest adding to it is a "Close" icon. Maybe an X in the lower right corner or something like that. The reason is that when using it in the search field for example in Radio view you could click off of it and land on a station which would start playing.

AlbertoNarduzzi:
Hi Tim,

indeed, this was the first idea, as some keyboards on mobile phones: they
open up automatically, but you close them, either by pressing ENTER or by a
specific close-button.
I'll implement it straight away.

P.S.
Don't forget there is only one text field not accepting the ENTER key
(because it has no onkey* event handler, and it relays on the form submit,
which cannot be fired by the OSK, of course.
I did fix it, but now it's lost... anyway I'll try to remember what I did,
it was like three lines of code.
Every single other text input field in the whole application works fine.

Cheers, Al.

PR #563, opened 2022-10-15
Added Artist to CoverView

AlbertoNarduzzi:
Made ARTIST and TRACK information independent in the CoverView.
Moved the cover-art slightly to the left.


moodeaudio:
After further testing there are a number of side effects of adding the ss-currentartist div that cause margin and spacing issues for the metadata in mobile landscape and wide layouts. It's mainly due to the existing code being designed around two div's of metadata instead of three.

It's best to revert these changes for now.

AlbertoNarduzzi:
Yep, I too saw a non-pleasant display in default mode. Wide mode, though,
looks ok (and I believe it does with any screen size / aspect-ratio).

I'll try to play a bit with it, and try to have my changes ONLY if in
cover-view-wide; leaving the cover-view default mode to display the
metadata as-is, like in the TAG and ALBUM views. See what I can come up
with...

PR #564, opened 2022-10-16
Use systemd automounts for mounting music sources

### LINKED ISSUE #545: Use systemd automounts instead of "manual" mounting for music sources
Currently it seems that configured music sources are mounted by using the `mount` command without any entry in `/etc/fstab`.

This can be a problem if the corresponding servers go away frequently (e.g. NAS goes to sleep) or the connection is unreliable. Such manual mounts break quite easily in these situations and in the worst case leave behind a non-working mountpoint, which does not respond and slows down everything that tries to access it. 

There is however a feature of systemd called automounts. [Configuring](https://www.freedesktop.org/software/systemd/man/systemd.automount.html) is quite easy and it takes care of mounting and unmounting shares whenever needed. I expect that by using it, the stability of those mounts would be greatly improved.

sll552:
fixes #545 

I eyeballed most of the things here as I have no real idea of the big picture yet. Especially the error handling is something I don't know how it's supposed to be done. Anyway I would appreciate if you could take a look, I tested it on my setup and it seems to work fine. 

I guess a point for discussion would be the migration of existing installations. A simple backup/restore of the config could work as the previous solution was not really persistent (meaning no extra files anywhere). But I haven't tested that.

What this whole change does is that it replaces the manual `mount` and `umount` calls with systemd mount/automount units. I didn't want to break any API so it's a little more complicated than it could be in the end.
If this gets approved it would be a good idea to remove most of the calls to `sourceMount('mount',....)` etc. because they shouldn't be necessary anymore as mounting happens automatically through systemd. When this is done the respective code in `music-source.php` could be cleaned up and removed.

### DIFF
--- a/www/daemon/mountmon.php
+++ b/www/daemon/mountmon.php
@@ -32,56 +32,28 @@ while (true) {
 
 	sleep(30);
 	$mounts = sqlRead('cfg_source', $dbh);
+	$exitCode = 1;
 	if ($mounts !== true) {
 		mountmonLog('mountmon: Checking remote mounts');
 		foreach ($mounts as $mp) {
-			// See if host is up
-			//mountmonLog('- Checking host ' . $mp['address'] . ' for mount ' . $mp['name']); // DEBUG
-			$result = sysCmd('ping -A -4 -c 1 ' .  $mp['address'] . ' 2>&1 | grep "Destination Host Unreachable\|Name or service not known"');
-			if (empty($result)) {
-				mountmonLog('- Host appears to be up');
-				// See if mount dir exists. It may not since umount/rmdir is done at shutdown/reboot
-				if (file_exists('/mnt/NAS/' . $mp['name'])) {
-					// See if file sharing service is accessible on the host
-					if ($mp['type'] == 'cifs') {
-						//mountmonLog('- Checking SMB mount'); // DEBUG
-						$result = sysCmd('ls /mnt/NAS/' . $mp['name'] .' 2>&1 | grep "Host is down\|Stale file handle"');
-						$fileSharingAccessible = !empty($result) ? false : true;
-					}
-					if ($mp['type'] == 'nfs') {
-						//mountmonLog('- Checking NFS mount'); // DEBUG
-						$port = '2049';
-						sysCmd('nmap -Pn -p ' . $port . ' ' . $mp['address'] . ' -oG /tmp/nmap.scan >/dev/null');
-						$result = sysCmd('cat /tmp/nmap.scan | grep "' . $port . '/open" | cut -f 1 | cut -d " " -f 2');
-						$fileSharingAccessible = !empty($result) ? true : false;
-					}
+			$unit = getUnitForMount($mp['name']);
+			$result = sysCmd("systemctl is-failed \"$unit\"", $exitCode);
+			// is-failed returns 0 if the unit is in failed state
+			$fileSharingAccessible = $exitCode == 0 ? false : true;
+			// Attempt remount
+			if ($fileSharingAccessible === false) {
+				$logs = sysCmd("journalctl -q -n 5 --no-pager --unit \"$unit\"");
+				$mp['error'] = "Mount error: \n".implode("\n", $logs);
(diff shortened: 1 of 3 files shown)


moodeaudio:
The challenge with this change is that the current method for mounts and monitoring generates very few if any trouble reports from users so doing a wholesale replacement does not seem to be necessary.

sll552:
Well it's not meant to be a fix for something (I mean it fixes one of my problems with moode), but more of an enhancement. In the end doing it this way would eliminate a lot of possible weak points in the current implementation. 
As I said it would be possible to eliminate almost every call to `sourceMount` in the whole codebase, also the `mountmon` can be largely simplified (and it's mere existence means that there are problems with the current approach). Currently there is basically no error handling, as it wasn't even possible to check the exit codes. Which means that it could have happened that a `umount` did not work and subsequent mounts would be mounted over the umounted but still existing mountpoint (at least when using lazy unmounting).
There is just so much stuff that could go wrong with this rather manual approach compared to using a standardized and proven way.

moodeaudio:
I don't disagree with what you are saying but as I mentioned earlier the current approach is not generating trouble reports and so I don't feel an urgent need to make integrating and testing a completely new mount mechanism priority 1 ATM.

I believe your approach is prolly the way to go but I'd first need to carve out some time to understand how the systemd automount mechanism works and how to troubleshoot it.

sll552:
> I don't disagree with what you are saying but as I mentioned earlier the current approach is not generating trouble reports and so I don't feel an urgent need to make integrating and testing a completely new mount mechanism priority 1 ATM.

I understand the "don't fix what's not broken" approach, but I believe that there are many instances of issues that most people just ignore or restart or press the remount button before taking the step to either create a forum or github account and posting their issues. And without much knowledge of linux (which I suspect most of the users do not have) it's hard to determine that such an issue could be easily solved or shouldn't even exist. 
My intention is that integrating the new approach should not result in much (if any) work for you. Testing of course would be nice and I believe it is worth it as in the end it simplifies things and actually removes the need to test the various mount/remount scenarios in the future. It's a draft for a reason, I am happy to add or rework stuff if it's not the way it should be, just let me know.
Also it's understandable that there are higher priority tasks, but this comes across as a nice way to say "Yeah, thats never going to happen".
Actually I am planning to rework how the "services" are currently managed, because right now it's some mixture of manually starting/stopping via systemctl and some stuff that is just started in the background and left on it's own. Integrating this would have the benefit of making it easier to depend on mounts. So in the end every service could be managed using systemd instead of manual intervention of the worker.

> I believe your approach is prolly the way to go but I'd first need to carve out some time to understand how the systemd automount mechanism works and how to troubleshoot it.

Well that is pretty easy. Systemd creates an autofs mount at the specified mountpoint and when something accesses that mointpoint it triggers the actual mount. The autofs part corresponds to the `.automount` unit which then triggers the `.mount` unit which does the mounting. Those units are more or less independent, which means that you could start/stop the `.mount` unit regardless if there is a `.automount` or not and have basically a mount/umount for this specific mountpoint.
Automount is just nice because it takes care of mounting based on file access which eliminates the need to make sure that everything is in place before something accesses the mount. As a bonus it also remounts stuff if it breaks.
Underneath is still the same mount mechanism so it's not a fix for everything, it can still fail and it can still hang, because that's how it is. But it's much easier to detect (just `systemctl status <mountpoint>`) if something bad happens and it won't break stuff in succession to a bad or failed mount (think of the double mount issue when using a wrong umount).
I don't really know what you mean by troubleshooting because it's a systemd unit, just like every service, so it has a status and logs which can be checked through `systemctl` and `journalctl`. The error messages are the ones from the actual mount command, so those will be the same as now. It's just easier to tell if the mount is there or not.

sll552:
I have added an actual improvement, the credentials are now stored in a separate file with more restrictive permissions which improves security.

moodeaudio:
Ok, thanks :-)

Still on the TODO list for testing at some point.

moodeaudio:
Closed due to lack of activity. Will reopen if needed.

PR #565, opened 2022-10-16
Metadata in coverview wide

AlbertoNarduzzi:
The CoverView artist is now rendered ONLY in WIDE mode; DEFAULT mode is exactly as it was.
Also, the whole metadata in WIDE mode is vertically centered.


moodeaudio:
I'll look into this after the CSS refactoring changes have been committed in a few weeks.

moodeaudio:
Al - I'm in the process of updating the layout of metadata in wide mode so stay tuned.

moodeaudio:
lol, time is relative so from my reference point it seems like it was just yesterday ;-)

PR #566, opened 2022-10-17
Cover view without playbar

AlbertoNarduzzi:
If we chose not to show the playbar in CoverView, this basically means there is more room for the cover-art itself, so just resized the img tag accordingly if body is :not(.cvpb); this improves the user experience for the extra-wide display.


moodeaudio:
I'll test after the CSS refactoring changes have been committed in a few weeks.

PR #568, opened 2022-10-26
Performed htmlspecialchars translation at mpd response level

AlbertoNarduzzi:
In order to have more coverage on special html characters - more or less intentionally put - in the audio-files metadata, moved the specific translations at mpd response level; this should guarantee no html re-interpretation on tag content.

Also, the html element showing the by-artist / by-advanced-search filtered library had its TEXT, and not its HTML, filled with the filter, causing the special characters to be shown as ampersand-mnemonics instead of their visual representation. FIXED.

### DIFF
--- a/www/command/audioinfo.php
+++ b/www/command/audioinfo.php
@@ -122,7 +122,7 @@ function parseTrackInfo($resp) {
 					$genres .= $value . ', ';
 					break;
 				case 'Album':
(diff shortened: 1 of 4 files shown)


moodeaudio:
The challenge with this change is that it adds some overhead to every MPD response even though it's quite rare to have HTML code in metadata tags.

This kind of edge case should be remedied by user manually cleaning up the tags.

PR #578, opened 2022-11-03
Album and disc differentiation

AlbertoNarduzzi:
The tracks comment / key was always included in the album name, preventing the correct retrieval of the tracks by album-name
The disc number was simply checked, with no additional filtering, causing inclusion of all the tracks with such a disc number, regardless of the album they belonged to.

Added plain album, comment, key and disc as attributes of the ALBUM and DISC headings to be available at the heading-click handler.

### DIFF
--- a/www/js/scripts-library.js
+++ b/www/js/scripts-library.js
@@ -795,7 +795,11 @@ var renderSongs = function(albumPos) {
             var album = filteredSongs[i].album + comment;
 
             if (album != lastAlbum) {
-                albumDiv = '<div class="lib-album-heading"><a class="btn" href="#notarget" data-toggle="context" data-target="#context-menu-lib-album-heading">' + album + '</a></div>';
+                albumDiv = '<div class="lib-album-heading"' +
+                        ' heading-album="' + filteredSongs[i].album + '"' +
+                        ' heading-comment="' + filteredSongs[i].comment + '"' +
+                        ' heading-key="' + filteredSongs[i].mb_albumid + '"' +
+                        '><a class="btn" href="#notarget" data-toggle="context" data-target="#context-menu-lib-album-heading">' + album + '</a></div>';
                 lastAlbum = album;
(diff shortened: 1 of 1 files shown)


moodeaudio:
How do I test this?

AlbertoNarduzzi:
Hi Tim,

prerequisite: you need an artist with multiple albums, and also some of
these albums must have multiple discs (maybe it's harder for you to do than
it is for me); then:
in the artists list, select just that artist, BUT DO NOT CHOSE ANY OF ITS
ALBUMS, so that the songs list is populated with all the tracks, along with
ALBUM and DISC headings.
Now, if you have at least two albums with disc1&2, and try to add disc1 of
album1 to the playing queue, you'll notice that also disc1 of album2 is
getting added, as there is no discrimination on the album name, at this
point (there will be, though, once you select an album from the albums
list).

Also, if you have the "include comment" or albumID selected, the album name
inside the heading is used for album match (when adding a whole album from
the songs list), but then, since such heading now also includes either the
comment, or the album id, so such album can be found...


Cheers, Al.

P.S. sorry for answering late...

moodeaudio:
Hi Al,

I’ll have to look through my test collection but do the files you sent me for cue testing meet the criterion you described for testing?

-Tim

AlbertoNarduzzi:
Hi Tim,

I believe yes: one of them has 2 discs, while the other has 3. Without my
fix, once you have selected Jimmy Van M, you can scroll down the songs
list, click on the disc 2 of the first album, and select ADD.
Then you can go to the playing queue and see...

moodeaudio:
Ok, looking good after testing.

It looks like the Album name heading should be highlighted similar to the disc heading when tapped. I can make that update.

AlbertoNarduzzi:
Yep.

I was busy doing the fix, and didn't even think af it; yes it's consistent.

PR #579, opened 2022-11-06
ID3v2 tag handling for DSD files

AlbertoNarduzzi:
Added an extension to the Zend library to extract the metadata address from a DSD file, and initialize a id3v2 class to read it.
NOTE:
New files added, as well as new directories.


moodeaudio:
Is this enhancement only to enable reading embedded cover art in DSF?
The reason I ask is that metadata tags are read ok in the DSF files that I use for testing.

AlbertoNarduzzi:
Yes, it is just to correct the behavior of the Zend library which assumes
the id3v2 tag is always at the beginning of the file.
MPD and mediainfo know it, therefore can (and do) extract all the other
metadata (artist / album / date / genre / etc.) correctly.

moodeaudio:
Merged but the PR didn't automatically close here so I'll close it manually.

AlbertoNarduzzi:
Yes,

I saw that you integrated it, but didn't show up as closed.
Glad it made it into moOde official. :-)

moodeaudio:
I'll prolly refactor inc/extensions files directly into the Zend dir for consistency.

AlbertoNarduzzi:
Hi Tim,

sorry, been busy...

Of course, as you see fit; I just made the things outside the main Zend
directory to keep them separated, as nothing being official.
We might need to check if all extensions the DSD format allows for are
contemplated; the id3v2 tag will always be addressed the same way, through
the DSD file header, so no change needed anywhere.

PR #591, opened 2023-01-04
Restructure nginx config

bitkeeper:
Restructue the config of nginx:
* Split the configuration of nginx over several reusable files, to make it easier to maintain and extend functionality
* Let nginx proxy the camillagui (so only one external visible http server is needed)
* Prepare some things to make the use of https easy, but don't activate anything.

Changes:
* Extract the site settings from the global `nginx.conf` to make it a global config (not moode specific)
* Add general template for the moode locations in `moode-locations.conf`
  * Includes proxy settings to make the camillagui part of the nginx server by using a proxy
* Added sites (moode specific) to `sites-available/`
  * added `moode-http.conf` for http
  * added `moode-https.config` for https
* Load site settings from `sites-enabled/*.conf`
  * Make symbolic link to  a site from` sites-availlable`

The symbolic link needs to be created from the moode-player package `postinstall.sh`. Also the default link needs to be removed. Will create separate PR in pkgbuild for it.

Also use of https is prepared for users who want to get their hands dirty.
It requires:
* Key of cert in `/etc/ssl/private/moode.key`
* Cert in `/etc/ssl/moode.cert`
* Switch the link in `/etc/nginx/sites-enabled` to `moode-https.conf`


moodeaudio:
Looks good.

PR #592, opened 2023-01-06
Fixed Analog and S/PDIF selectable as Input

diocloid:
Analog and S/PDIF were not selectable due to the pre-refactor array key being used.


moodeaudio:
Thanks. I'll merge later this morning.

PR #594, opened 2023-01-13
Size of the album-art was not pefect

AlbertoNarduzzi:
To see the meaning of this fix you need to watch the album art in the TAG view:

chose an artist with more than one album (don't select any album, though, just the artist in the list on the left) and see how wide(r) is the faded album art underneath the Artist's name text overlay.
now select an album from the list on the right, and notice the album art is (slightly, but still noticeably, to me) smaller
To overcome I made the following:

removed position: Absolute from the IMG tag, and assigned, instead, to the button overlay, and positioned it exactly OVER the IMG
2.To make such button work correctly also when no artist is selected ("Music Collection" scenario) I also added an IMG tag with the imagesw/notfound.jpg
This problem is visible ONLY with the 7'' display, and very, very slightly on my iPhone (7). On the browser, OTOH, is absolutely absent, as the image tag's maxwidth: 100%, takes care of it...

I know it's just cosmetics, and not for everyone, but I am a perfectionist when it comes to graphics; and since I use almost ONLY the 7'' display, I feel frustrated having to live with the little misalignment... ;-)

Of course, it's always up to you whether to include it or not in the upcoming 8.2.5

### DIFF
--- a/www/css/media.css
+++ b/www/css/media.css
@@ -334,7 +334,7 @@ body {
 	#trackscontainer span {font-size:1.1rem;}
 	#lib-coverart-meta-area {width:14vw;position:relative;}
 	#songsList {padding-bottom:10rem;}
-	#tagview-text-cover {font-size:1.35rem;height:calc(14vw - 1rem);width:calc(14vw - 1rem);}
+	#tagview-text-cover {font-size:1.35rem;height:calc(15vw - 1em);width:calc(15vw - 1em);}
 	img.lib-artistart {width:inherit;height:inherit;}
 	.lib-numtracks-meta {margin-top:1.5em;}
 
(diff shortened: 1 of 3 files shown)


moodeaudio:
I'm able to repro at height <= 479px so prolly a bug in the general landscape mode media query.
I'll investigate.

moodeaudio:
This change to only media.css seems to do the trick.
```
@media (max-height:479px) and (orientation:landscape) {
.
.
	#tagview-text-cover {font-size:1.35rem;}
	#tagview-text-cover,
	img.lib-artistart {height:calc(14vw - 1rem);width:calc(14vw - 1rem);}
```

Give it a try and let me know.

moodeaudio:
Closed due to alternate way of fixing the issue as described in previous post plus subsequent modification proposed by @AlbertoNarduzzi to change from rem to em for precise positioning.

PR #599, opened 2023-01-23
Feature/camilladsp volume

bitkeeper:
This PR will make it possible to sync CamillaDSP volume to the current MPD volume.

Benefit:

- Can use CamillaDSP volume filter (+ dither is needed)
- Is that it will be later in the audio chain then using the current software volume of MPD.
- And because the volume silder of CamillaDSP is used also the Loudness filter can be used.

The implementation makes use of [mpd2cdspvolume](https://github.com/bitkeeper/mpd2cdspvolume)
On the moode side an additional MPD Volume Type 'CamillaDSP' will be avialable at the Audio configuration page.
Only available when CamillaDSP is used with moOde (=configuration selected).

When the CamillaDSP volume type is selected the following is happing:

- In the background MPD mixer type 'null' is used.
- The mpd2cdspvolume service is started

### DIFF
--- a/etc/alsa/conf.d/camilladsp.overwrite.conf
+++ b/etc/alsa/conf.d/camilladsp.overwrite.conf
@@ -4,6 +4,10 @@ pcm.camilladsp {
     cpath "/usr/local/bin/camilladsp"
     config_out "/usr/share/camilladsp/working_config.yml"
 
+    # use a state file for restoring last volume settings on start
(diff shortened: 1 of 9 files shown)


moodeaudio:
To init the volume state file
`echo "0 0" | sudo tee -a /var/lib/cdsp/camilladsp_volume_state`

PR #634, opened 2023-07-08
Update README.md with developer info.

janiversen:
First PR, more to follow promptly.

Updated README, with some of the comments from the forum, and a couple of items I think was missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,30 @@ Tim Curtis © 2014
 [Setup guide: ./www/setup.txt](./www/setup.txt)<br/>
 [Contributors: ./www/CONTRIBS.html](./www/CONTRIBS.html)<br/>
 
+### Solve problems
+
+Please check our [FAQ and guides](https://moodeaudio.org/forum/forumdisplay.php?fid=9) section of the forum,
+before filing an issue.<br/>
+
+Please check our [Support](https://moodeaudio.org/forum/forumdisplay.php?fid=7) section of the forum, if you need support.<br/>
+
+If you have found a bug, then please report it using [github issues](https://github.com/moode-player/docs/issues)<br/>
+
+### Contribute to the project
+
+Contributions are welcome.
+
+The [Master Development Guide](https://github.com/moode-player/docs/blob/main/master_development%20guide.md)
+contains information about how to setup/develop/test moode audio as well as an explanation of the architecture<br/>
+
+Please clone the repository, make a feature branch based on branch "development".<br/>
+Develop and test the feature<br/>
+Once your work is completed, please submit a [Pull Request](https://github.com/moode-player/moode/pulls)<br/>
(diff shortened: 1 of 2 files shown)


moodeaudio:
Couple of things.

1. We definitely don't want to maintain a public TODO list. 
- The Forum already contains a thread for each upcoming release that provides WIP release notes describing whats included in the release and sometimes a note about what might be in the release.
- I maintain a personal check list to help me keep track of what I'm working on and this is what I refer to when I post in the Forum "Added to the TODO list".
- If you are looking for information on some TODO's with respect to the code base itself for example Refactoring, I'd suggest making a new Thread in the Forum and titled "Code base maintenance" or something like that.

2. For the readme.md page the challenge with adding to it is that 99% of the time this page is only viewed by people with GitHub accounts which means that they are most likely devs and so don't need any info on how to clone repos and make PR's.
- It would be helpful to add a "Troubleshooting and Support" heading with some text explaining that this is provided in the Forum (link to Support section) and not via Git issues which should only be used for development related stuff.
- Also helpful would be a link to the docs repo under the "Additional Resources" heading.

janiversen:
The TODO is a matter of taste, and no problem. To me it is a convenient way of sharing information.

Actually a lot of new developers in projects needs hand holding to make the first steps. But please not I did not add "git clone" specifically, I added a section that contributions are welcome, and how does the projects wants bugs/enhancements to be handled.

I have worked quite long time mentoring newbies in projects like Libreoffice and organizations like Apache, so I think I know about about what people look for in a readme.

However it is your project, and you decide to merge or not.

janiversen:
I added specifically "solve problems" (title can be changed) to tell how to get support in contrast to reporting a bug. I at least would never report a bug in a forum, nor would ask about support in a issue.

janiversen:
Sorry for disturbing, I was honestly just trying to help, by bringing my experience to the table.

moodeaudio:
There no problem with the PR and it's a helpful one. I'll update readme with some of the suggestions.
I suppose devs new to the project would need some "getting started" type info.

janiversen:
Well your comments was quite different, it sounded a lot more like my PR was a burden, something I am very observant to, when starting in a new project.

I guess we view the project from different angles. As a newbie but with more project experience than most, I see a lot of issues, that are clear to the people who have been around a while.

I saw a possibility to combine developing the features I need and simultaneously help the project make a bigger step forward (by doing things the project need), mostly because I basically have all the time I want. When I engage in a project it is because there are mutual benefits, life is simply too short to fight an uphill battle.


This is open source so feel free to use what I suggested.

moodeaudio:
No burden at all and the help is most welcome.
I prefer to get my perspectives on the table to stimulate a meaningful discussion. They are not by any means set in stone.
Often the proposals in PR's that are focussed on content/style rather than bug fixes take a lot more more time to consider.

PR #635, opened 2023-07-08
Amend setup.txt with some tips.

janiversen:
Added some tips, among other the tip on how to find a real radio stream.

I think these changes are relevant at user level.

### DIFF
--- a/www/setup.txt
+++ b/www/setup.txt
@@ -190,6 +190,10 @@ PLAYER SETUP AND CONFIGURATON
 
 At this point a FULLY OPERATIONAL PLAYER exists.
 
+FURTHER  HELP
+
+is available in our forum, espiecially [FAQ and Guides](https://moodeaudio.org/forum/forumdisplay.php?fid=9)
+
 CUSTOM CONFIGURATIONS
 
 Customize the player by using any of the following procedures.
@@ -263,6 +267,8 @@ Follow these instructions for making certain types of changes.
    e) Remove WiFi adapter
    f) Power on
 
+
+
 COMMAND AND REST API
 
 Commands can be submitted to moOde via HTTP or SSH. If an HTTP command returns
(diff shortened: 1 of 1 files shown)


TheOldPresbyope:
There may be indirection or not. Depending on the source of the link, the endpoint may be a playable stream, it may be a .pls file, sometimes it may like the case explicitly shown here. Over the years we've seen lots of different links offered up by various users.

janiversen:
Agreed, but this at least describes one way to get a proper stream....it solved a problem for me, and just maybe it can solve problems for others as well.

I have no problem adding extra text or changing the text. But I do think it is a nice hint on how to get the stream, so not writing anything just because it does not cover 100% seems strange to me.

Do you have a proposal on how to make the text to your liking ?

moodeaudio:
It's useful information but it should be in the "FAQ and Guides" sub-forum and not in the Setup Guide.

The Setup Guide is meant to contain the steps needed to get a player operational out of the box plus some information on the command API.

janiversen:
Ok I will move it, and add a reference to that sub forum. The problem I see is finding the information without knowing where it is,

janiversen:
Content is moved, but the least we should have is a pointer to the forum, as newbies might not know it exist.

WDYT

moodeaudio:
The challenge is that its human nature to "not read the manual" but in any case the Support and Developer pages at moodeaudio.org contain all the info and links needed to get to going.

janiversen:
OK, I get the message!

I will close the PRs as they are not helpful to me, but could be to others. Sorry for disturbing. 

I am happily developing the feature I need, it turned out that what I need is a "special' multi room client, allowing me not only to use my earbuds, but also to use the speakers on my mac. I will post in the forum when it is ready to be used by others.

moodeaudio:
The multi-room client sounds cool.

PR #636, opened 2023-07-10
Set volume at 100% to 0dB instead of +4dB for HDMI 1 and HDMI 2

JWahle:
This was implemented before (See #275 and #298) but broke due to mixer name change from "HDMI" to "PCM" for Raspberry PI HDMI devices.
I assume this happened, when the 2nd HDMI port was introduced for the Raspberry Pi 4.

I only own a Raspberry 3b+ and 4.
The bug was present on both and the fix works on both.
The device names `Pi HDMI 1` and `Pi HDMI 2` are used throughout the Moode codebase, so I assume they can be used reliably.
This fix should continue to work, even if the Raspberry Pi foundation changes the name of the HDMI mixers or sound devices, so this implementation should be future proof.

### DIFF
--- a/www/util/sysutil.sh
+++ b/www/util/sysutil.sh
@@ -76,9 +76,9 @@ if [[ $1 = "get-alsavol" || $1 = "set-alsavol" ]]; then
 		exit
 	else
 		# Set-alsavol (Note: % is appended to LEVEL)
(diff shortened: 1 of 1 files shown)


moodeaudio:
Nice :-) 

I'll merge and test tonight.

JWahle:
Very nice, thanks :+1:

PR #640, opened 2023-08-20
Added channel filters

AlbertoNarduzzi:


moodeaudio:
The challenge with this one is that there prolly aren't many multichannel tracks (if any) in most collections and so the filters won't get used much but the check code in genLibrary() loop gets executed for every track. Also the proposed display format is kind of confusing for example "16/44.1/2" and there is a wrap/width issue when "badge" is selected as the Pref.

See attached screen shot mock-ups.

#### Playback and CoverView
Screenshot 2023-08-22 at 7 33 30 AM
Screenshot 2023-08-22 at 7 42 48 AM

#### Library (Rubber Soul Album)

Screenshot 2023-08-22 at 11 35 47 AM
Screenshot 2023-08-22 at 11 36 37 AM

AlbertoNarduzzi:
Hi Tim,

I have chosen to add the number of channels there (it kind of belongs to there) for not adding a separate item in the audio info window (in the currently playing there is such information, but not in the file's), and actually having a filter without a field for showing how many channels, seemed silly to me.

Although I do agree that the multi-channel files are a minority (I for one have only 3 albums out of 1000+), I still find such info useful... (as the fix for the separate album cover + single-file embedded art of the NIN The Slip...)

Anyway, you decide; if we can go for adding a separate entry for the channels, and hence remove them from the format-string I am still OK; the most important thing being the relevant filters of course.

Cheers, Al.

moodeaudio:
"Channels N" and "Multichannel" could be useful filters if only for the sake of completeness but I'm still concerned about adding another rarely used filter check to the genLibrary loop. 

I'm also of the mind that displaying channel info is only practical where there is space to do so for example in Audio info (Track/Playback) like below. This much text in Extra Metadata in Playback/CoverView would be just TMI (too much information).

16 bit, 44.1 kHz, Stereo FLAC
16 bit, 44.1 kHz, 5-Channel FLAC

In Album view the default compact form w/o channels needs to be used because of space constraints. I don't see a way around that.

Overall the thing I want to avoid is having the UI look like the display on an AVR :-0

moodeaudio:
Yes, go ahead with new PR.

AlbertoNarduzzi:
Closing this PR, as everything - correct - is in Multichannel3, both channel display and filters

AlbertoNarduzzi:
Hi Tim,

I closed the previous PR and made a new one with channel information
available only inside audio-info window; channel filters as were.
NOTE: the *encodedAt* string returned by music-library.php is *still
including the channel* number, *but it's* then *recomposed* as needed *for
display*...

Cheers, Al.

PR #641, opened 2023-08-24
Multichannel3

AlbertoNarduzzi:
getEncodedAt now returns bit / sample / channels / filetype
this string gets exploded and used in:
Audio Format => [0] / [1]  [3] (the string remains the same)
Channels => [2] (added audio info line for this)

Checked, outside the audio-info window everything is showing as it has always been


moodeaudio:
Zip up one of the multichannel albums or  just a few tracks and email me a download link so I have something to use for testing. Thx.

PR #647, opened 2023-12-18
Added disc and track to the current track check

AlbertoNarduzzi:
The cycle sets the icon class, but since it is generating the list it does not break (like in the playback queue), so two equal titles of the same album always got marked.
I added the check for tracknum and disc. (there's only one place where it is set in the tag view, and that's when the list is being generated)

Actually, IMO replacing the title with the track number should be enough; but including also the disc number makes it work for separate CUE discs, where each disc begins with track number 1.

P.S.
I swear I have seen such icon also in the folder view; although cannot find any code that sets it... I'm probably dreaming of it...

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -4663,7 +4663,7 @@ function setNpIcon() {
         $('#songsList .lib-entry-song .songtrack').removeClass('lib-track-npicon');
         if (MPD.json['artist'] != 'Radio station' && $('#songsList li').length > 0) {
             for (i = 0; i < filteredSongs.length; i++) {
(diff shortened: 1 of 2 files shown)


moodeaudio:
I think my attempt to repro the issue with your test files is not working correctly. I emailed you a screen shot of tag view.

AlbertoNarduzzi:
Yes, seen it.
Anyway, I will add one more fix, so wait for my confirmation before
attempting to integrate the PR

moodeaudio:
I need test files that allow me to repro the issue before doing anything with the PR

moodeaudio:
Nvm, needed to turn off Ignore CUE in Lib Config

PR #655, opened 2024-02-19
Fix recreation of currentsong.txt

akelge:
Currenty currentsong.txt is updated if some fields changed, but the comparison `$fileMeta['outrate'] != $hwParamsFormat . $hwParamsCalcrate)` will always fail, cause actually in the file we write `$current['output'] . $hwParamsCalcrate . "\n";`.
This way the file is updated even when the player is in `stop` or `pause` mode.

This patch fixes the comparison and adds a comparison `$fileMeta['bitrate'] != $current['bitrate']` to trigger an update of currentsong.txt in case the bitrate changes.


akelge:
Included in the 8.3.9 release

PR #669, opened 2024-07-18
getUserID - fix for in-place update

fabitom:
To avoid misunderstandings and misinterpretation of the entry 'pi,' we should look for the directory name instead of the word 'pi'.


moodeaudio:
Yes, good catch. I'll merge in a bit.
Thx

PR #670, opened 2024-07-19
Added 24-hour digital clock

fabitom:
24-hour format added to the digital clock of the screensaver


moodeaudio:
Good idea but a simpler implementation would be to use the existing SQL param "scnsaver_mode" and add a new value "Digital clock (24 hour)" for example

```
... data-cmd="scnsaver-mode-sel"><span class="text">Digital clock</span></a></li>
... data-cmd="scnsaver-mode-sel"><span class="text">Digital clock (24 hour)</span></a></li>
```
```
function showSSClock() {
	switch (SESSION.json['scnsaver_mode']) {
		case 'Digital clock':
		case 'Digital clock (24 hour)':
                        var showAMPM = SESSION.json['scnsaver_mode'] == 'Digital clock (24 hour)' ? true : false;
			showSSDigitalClock(showAMPM);
			GLOBAL.ssClockIntervalID = setInterval(showSSDigitalClock, 1000);
			break;
```

fabitom:
Yes, it may indeed be much simpler, simplified.

moodeaudio:
Theres something odd happening in Git Desktop on my end when I try to merge the PR. It's saying the the develop branch is already up to date with the PR even though this is not the case.

If not too much trouble can you delete this PR and make a new one? Thanks.

fabitom:
Yes, of course.

PR #673, opened 2024-07-20
fix for update clock - setInterval

fabitom:
incorrectly called function in setInterval


fabitom:
Oh, you've already fixed it

Issue #671, opened 2024-07-19
Timed CoverView display during playback


fabitom:
Hey! It's possible that I don't fully understand the idea of the coverview/screensaver, especially on a local LCD, but I think there should still be an option for the screensaver to work during playback, since we have a playbar that can be shown on the screensaver.
From the description, it seems it will only be active when MPD is not playing and the renderers are deactivated.
Currently, I'm testing unlocking the screensaver and I don't see any issues with the screensaver during playback.
I can make a PR if you want.

moodeaudio:
CoverView usage scenarios:
1. Timed display
This is classic screen saver mode. It's automatically activated after the timeout period when there is no playback activity (MPD or renderers). It's typically displayed w/o the Playbar.
2. Automatic display
This option supports the usage scenario where the Pi is connected to a large display at the front of the listening area, usually in-between the speakers/audio gear, and playback is controlled by Smartphone/Tablet at the listening chair. Timed display would be set to Never and Playbar set to No in this case.
3. Alternate main Playback view
This is supported by clicking on the TV icon. Timed display would be set to Never and Playbar set to Yes in this case. MPD can be playing. Timed display could also be set to a value but there has to be no MPD playback activity for it to take effect.

When a renderer connects its overlay is always displayed to prevent so user has access to the Disconnect/Turn off buttons. Access to the main WebUI is blocked to prevent changing any settings that would break the renderer config.

I think you are proposing a modification to Timed display whereby it would activate even if MPD is playing, correct?

fabitom:
Yes, exactly. I often listen to music before and during sleep ;) so having the screensaver with a clock left on overnight would be a nice option.

moodeaudio:
Will reopen if there is new information posted

PR #674, opened 2024-07-20
Allow Timed display when MPD playing

fabitom:
In reference to issue #671

### DIFF
--- a/www/relnotes.txt
+++ b/www/relnotes.txt
@@ -23,6 +23,7 @@ Updates
 - UPD: Add option to downgrade chromium-browser
 - UPD: Add link to renderer Setup guide on Renderer Config
 - UPD: Add stats for log2ram option on System Config
+- UPD: Add Prefs option to enable Timed CoverView when MPD is playing
 - UPD: Improve tracks section for mobile Tag/Album views
 - UPD: Improve Spotify Connect event handling
 - UPD: Improve vol.sh to unmute when volume is changed


moodeaudio:
We can't just delete the MPD test. It was added a long while back based on a lot of users not wanting CoverView to activate when something was playing.

This would have to be a new Prefs option but that involves a lot of moving parts and more code so I'm reluctant to consider it unless there was a sufficient demand for it.

What might be possible is to add a dev tweak, something like checking for the presence of a file in home dir. Something like the following:
`&& ($mpdState != 'play' || file_exists($_SESSION['home_dir'] . '/coverview_mpd_override') )`

fabitom:
Sure, I understand. It mainly depends on the primary use of the streamer — whether it is meant for listening sessions, daily radio, or as an alarm clock ;). Is there already a mechanism in place that forces the overwriting of certain settings not available in the interface options, or would this be the first use?

moodeaudio:
The php scripts would be overwritten during in-place update but the file in home dir would not. All you would need to do is make a sed patch that adds back the fie_exists() code.

moodeaudio:
Will reopen if there is new information posted

moodeaudio:
Update: 

I've added a Prefs option to control whether CoverView activates when MPD is playing.

Screenshot 2024-10-13 at 11 44 37 AM

fabitom:
Nice, thanks!

PR #712, opened 2025-04-28
[Bluetooth] Handle speaker disconnect/connect

poupounetjoyeux:
Using MusicAssistant, if the renderer raise an error, all the MA provider is in failure
This means that when streaming to a BT speaker, if the speaker is not connected, we need to restart MA 

To fix this I propose in this PR to bind renderers start/stop regarding the bluetooth connection status
Because we don't want to see in our client the speaker if it's finally not available
To make it working you will need to reconnect your speaker the firstime to make the MAC available inside SESSION

Feel free for any comment! 👍

### DIFF
--- a/www/blu-config.php
+++ b/www/blu-config.php
@@ -49,10 +49,16 @@ if (isset($_POST['rm_paired_device']) && $_POST['rm_paired_device'] == '1') {
 	$cmd = '-p';
 	sleep(1);
 }
+
+function updateBluetoothSpeakerMACAddress($deviceMac) {
+	sysCmd("sed -i '/device/c\device \"" . $deviceMac . "\"' " . ALSA_PLUGIN_PATH . '/btstream.conf');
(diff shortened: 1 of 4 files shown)


moodeaudio:
I have no idea what "Music Assistant" is but the Bluetooth feature in moode when its configured for "Bluetooth Speaker" supports audio sources including MPD, AIrPlay, Spotify Connect, Deezer Connect and Squeezelite.

poupounetjoyeux:
Music assistant is just an example
I think that the issue can occures also on mobile client and other streamers
The issue is more that moode raise an error if the bluetooth speaker is off
So why making it available on streamers if we can't stream on it ?

moodeaudio:
There are no reports in our support Forum of this type of Bluetooth issue. Your best bet would be to get a Forum account, post the issue and then see what kind of feedback you get.

For info on how to easily get a Forum account visit the Support page at moodeaudio.org.

PR #720, opened 2025-09-20
add: radio-station: DLF / DLF-Kultur / DFL-Nova

paepckehh:
Proposoal:
add DLF / DLF-Kultur / DLF-Nova Radio to distributed moode core radio station (sql.insert/url/hq-images)

Rationale:
- DLF provides [high quality streams in various open source formats ](https://www.deutschlandfunk.de/livestream-100.html)
- DLF provides advertisement free program (publicly financed radio, in-depended, open, innovative)
- DLF provides a high quality music and information program

With the progressive de-funding and closure of Public Founded Radio as in-depended information source (NPR, VoA,...) across the western democracies, please support and anchor this as viable independent information source. 

(DLF wiki)[https://en.wikipedia.org/wiki/Deutschlandradio] 

Side Notes:
- Started this as proposal instead of a Forum post, b/c I was unable to maintain a forum account for forum proposal/discussion.
- Not sure about how to proceed with station numbering, please see this PR mainly as convenient copy & paste source.


moodeaudio:
It doesn't seem like the stations are music-centric which would be more generally accessible for users. For example like FluxFM.

I'd suggest posing in our Forum to see what the interest level might be and so that users can add the links if they desire.

PR #740, opened 2026-04-03
Update playlist.php

slowm0e:
M3U parser ignoring #EXTINF titles in getPlaylistContents() #EXTINF and other standard M3U comment lines were being passed to MPD's lsinfo, returning no tags and displaying as "Unknown title" in the Edit Playlist view, and only showed MP3 adress in Playing. This also caused slowdown as MPD attempted to resolve each comment line as a stream. #EXTINF titles are now parsed and applied to the following URL entry. All other unrecognised # lines are skipped. No change to behaviour for song file playlists or radio stations.

### DIFF
--- a/www/command/playlist.php
+++ b/www/command/playlist.php
@@ -192,50 +192,60 @@ function getPlaylists() {
 }
 // Return playlist metadata and items
 function getPlaylistContents($plName) {
-	$plFile = MPD_PLAYLIST_ROOT . $plName . '.m3u';
-	$dbh = sqlConnect();
-	$sock = getMpdSock('command/playlist.php');
-
-	$genre = '';
-	$cover = 'default';
-	$items = array();
-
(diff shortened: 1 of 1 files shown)


moodeaudio:
The current playlist parser assumes an MPD style M3U file (just song file and station URI's) with the two additional custom moode header EXT's added for Genre and Cover.

Whats the usage scenario for m3u's with additional EXT tags?

slowm0e:
The use case is anyone creating an M3U playlist with standard #EXTINF titles — my specific case being podcast episode URLs where I wanted meaningful titles instead of raw URLs. But it would apply equally to anyone manually curating a playlist with custom titles for streams or any other HTTP content.
I can maintain my own patched playlist.php locally, but since this is standard M3U behaviour I figured others would eventually hit the same wall. The fix is small and doesn't affect existing song file playlists or radio stations at all.

PR #747, opened 2026-06-07
Add condition to check w_queue in waitWorker

HJKLMN:
Opening Configure from the main player UI and clicking any option (System, Renderers, Network, etc.) causes the browser to hang indefinitely. The page never loads. Navigating directly to the URL (e.g.  http://moode.local/sys-config.php ) works fine — the problem only occurs when clicking from the Configure modal while the player is active.

To fix is , i modified /var/www/inc/common.php modified waitWorker() to skip the wait when the only queued job is reset_screen_saver.

### DIFF
--- a/www/inc/common.php
+++ b/www/inc/common.php
@@ -565,7 +565,7 @@ function waitWorker($caller) {
 	//debugLog('waitWorker(): Start ' . $caller . ', w_active=' . $_SESSION['w_active']);
 	$loopCnt = 0;
 
-	if ($_SESSION['w_active'] == 1) {
+	if ($_SESSION['w_active'] == 1 && $_SESSION['w_queue'] != 'reset_screen_saver') {
 		do {
 			usleep(WAITWORKER_SLEEP);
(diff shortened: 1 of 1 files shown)


moodeaudio:
Duplicate of this Forum post
https://moodeaudio.org/forum/showthread.php?tid=8748

I'm not able to repro this on any of my systems.

moodeaudio:
I was able to repro something similar to your issue while investigating another issue.

Did you have the metadata file turned on?

HJKLMN:
Yes

moodeaudio:
This very old  bug involves the PHP session being prematurely closed by the metadata file routine in worker.php which causes waitworker() routine to hang indefinitely because the $_SESSION variables are not available.

Fixed for upcoming 10.2.3

HJKLMN:
Great! So this could happen other  places as well, I removed my filter fix, and tried to fix it in session.php.  As soon as your new version is available i will use your fix

PR #731, opened 2026-01-16
add (experimental) support for Sendspin Multiroom Audio Client Renderer

m-mrks:
First attempt to include Sendspin as renderer in moOde.

See the `setup_3rdparty_sendspin.txt` file for additional details on installing [sendspin-cli](https://github.com/Sendspin/sendspin-cli) in moOde; in short:

```bash
sudo pip install uv --break-system-packages
sudo uv tool install sendspin@4.0.0
```

To test via SSH if the Sendspin server can be reached; and desired audio device is available:

```bash
sudo /root/.local/share/uv/tools/sendspin/bin/sendspin --list-servers
sudo /root/.local/share/uv/tools/sendspin/bin/sendspin --list-audio-devices
```

Probably overlooked some basics, but am happy to learn / improve / collaborate :-)

PR #748, opened 2026-06-09
add dmix output mode option

m-mrks:
Add [dmix](https://www.alsa-project.org/alsa-doc/alsa-lib/pcm_plugins.html#pcm_plugins_dmix) plug type as ALSA audio output.

image


moodeaudio:
Whats the usage scenario where dmix would be useful?

m-mrks:
> Whats the usage scenario where dmix would be useful?

Allow multiple sources playing simultaneously without [device busy](https://github.com/moode-player/moode/pull/731#issuecomment-3978233999) errors.

moodeaudio:
That just explains what dmix can be used for.

Whats a real world scenario where users would have multiple music sources connected and mixed for playback?

PR #758, opened 2026-06-28
Fix Bluetooth A2DP sink silent in Standard ALSA output mode

Gjuju:
Be aware, huge description for a small correction/improvement !

Tested  with success on my Pi and x86 port moOde 10.2.4 with my 2 DACs, but other testers are welcome.

## Summary

When a phone streams to moOde over Bluetooth (moOde acting as the A2DP sink /
"Bluetooth speaker"), with Bluetooth mode = **Standard**:

- **The defect occurs only when the main ALSA output mode is `Direct` (`hw`) or
  `IEC958`** (the "raw" modes that do not convert formats), **and** no DSP and
  no Peppy are active. In that case there is **no sound**.
- **Enabling any DSP (CamillaDSP, Graphic/Parametric EQ, Crossfeed) or Peppy
  already fixes it** — those plugins insert a `plug` into the chain, which
  performs the missing format conversion. **Compatibility** mode also produces
  sound, but it bypasses DSP and Peppy entirely.
- On output mode `Default` (`plughw`) the defect never appears, because
  `plughw` itself inserts a `plug`.

In other words, the only broken path is plain Standard (no DSP/Peppy) on a raw
output mode — exactly the path that has no `plug` to convert the fixed-format
Bluetooth stream.

This PR generalizes that same conversion to the plain Standard path by routing
the inbound stream through `plug:_audioout` instead of bare `_audioout`. Result:
Standard plays in **every** configuration (with or without DSP/Peppy), DSP and
Peppy still apply when active, the MPD playback path is **untouched**, and
**bit-perfect output is preserved**.

Adding a software `plug` layer to the Bluetooth path is acceptable by design:
A2DP is inherently non-audiophile, and the lossy Bluetooth codec
(SBC/AAC/aptX/LDAC) already alters the signal far more than an ALSA format/rate
conversion ever would. The `plug` is therefore negligible next to the codec it
sits behind — and it only ever touches the Bluetooth path, never the lossless
MPD playback chain.

## Root cause

`bluealsa-aplay` emits a **fixed-format** PCM (the decoded A2DP stream, e.g.
`S16_LE 44.1 kHz` from SBC/aptX-HD). In Standard mode it is sent to `_audioout`:

    pcm._audioout { type copy; slave.pcm "<device>" }

`type copy` performs **no format/rate conversion** — producer and slave must
agree. moOde's own ALSA output mode help already describes the mechanism:

> **Direct**: ALSA "hw" plugin which **does not perform format conversions**. The audio device must accept the given format.
> **Default**: ALSA "plughw" plugin which **performs format conversions if needed** to match audio device requirements.

So in Standard mode, when the slave resolves to a raw device (output mode
**Direct/`hw`** or **IEC958**) and nothing in the chain converts, the DAC
rejects the stream's fixed format and the device fails to open → **silence**.

Two sub-cases are affected:

1. **No DSP and no Peppy** → `copy → hw`
2. **Invert polarity only** → `copy → invpolarity (type route) → hw`
   (`type route` does not convert either)

**Default (`plughw`)** already worked because `plughw` == `plug` + `hw`, which
inserts a converter. This is why the defect went unnoticed — most users are on
Default; it bites bit-perfect (`Direct`/`hw`) and IEC958 setups.

Compatibility "works" only because it points `bluealsa-aplay` directly at
`plughw:<card>,0`, attacking the card and therefore **skipping the whole
`_audioout` chain (no DSP, no Peppy)**.

## Why DSP/Peppy already worked (and still do)

These targets already carry their own conversion, so they were never affected
and are not changed by this PR:

| Target        | ALSA type                        | Converts? |
|---------------|----------------------------------|-----------|
| `peppy`       | `type plug`                      | yes (built-in) |
| `alsaequal`   | `type plug`                      | yes (built-in) |
| `eqfa12p`     | `type plug`                      | yes (built-in) |
| `crossfeed`   | `type plug`                      | yes (built-in) |
| `camilladsp`  | `type cdsp` (rate list, `-f/-r`) | yes (self-adapting) |
| `invpolarity` | `type route`                     | no |
| `_audioout`   | `type copy`                      | no |

## The fix

Prefix the Standard (`_audioout`) target with `plug:` so the conversion happens
**at the top** of the chain — between `bluealsa-aplay` and `_audioout` —
instead of mutating the shared `_audioout` slave:

    bluealsa-aplay -> plug -> _audioout (copy) -> [DSP/Peppy/invpolarity] -> hw -> DAC
                      \ converts the fixed BT format, negotiated down to the device /

- The `plug` affects only the Bluetooth producer.
- The chain still flows through `_audioout`, so **DSP and Peppy keep applying**.
- For targets that already convert (DSP/Peppy), the extra top `plug` is a
  transparent **no-op** (it only converts when the formats differ).

Mutating `_audioout`'s slave to `plughw` instead was deliberately avoided:
`_audioout` is shared with MPD, and that would insert `plug` into MPD's path,
breaking bit-perfect playback. The top-level wrapper avoids this entirely.

## Why there are no expected regressions

- **MPD playback path is untouched** — it still uses `_audioout -> hw` directly;
  bit-perfect output is preserved.
- **Bluetooth audio is lossy** (SBC/AAC/aptX/LDAC), so adding `plug` to the BT
  path has no audible/bit-perfect cost.
- **`plug` only converts when needed** and **applies no gain** — it is a
  pass-through when the producer format already matches the slave. The inbound
  Bluetooth routing is therefore independent of the MPD volume type
  (Hardware / Software / Fixed).
- **Compatibility mode is unchanged** and remains available.
- The default seed (`etc/bluealsaaplay.conf`) is updated so a fresh install
  works out of the box.

## Behavior matrix

Legend: silent / sound / DSP-Peppy applied / DSP-Peppy bypassed.

**Main output mode = Direct (`hw`):**

| DSP/Peppy | Mode | AUDIODEV | Before | After | DSP/Peppy |
|---|---|---|---|---|---|
| none | Standard | `plug:_audioout` | silent | sound | - |
| none | Compatibility | `plughw:N,0` | sound | sound (same) | - |
| Peppy | Standard | `peppy` | sound | sound (same) | applied |
| CamillaDSP | Standard | `plug:_audioout` | sound | sound (no-op) | applied |
| EQ / Crossfeed | Standard | `plug:_audioout` | sound | sound (no-op) | applied |
| Invert polarity only | Standard | `plug:_audioout` | silent | sound | applied |
| any DSP/Peppy | Compatibility | `plughw:N,0` | sound | sound (same) | bypassed |

**Only the two raw-`_audioout` cases change** (silent -> sound). Everything else
is unchanged or a no-op.

**Output mode = Default (`plughw`):** never affected — `_audioout`'s slave is
`plughw` which already converts; the fix is a no-op there.

**Output mode = IEC958:** Standard goes silent -> sound (the `plug` converts to
the IEC958 subframe format). Note a pre-existing, separate gap not addressed
here: **Compatibility + IEC958** uses a raw IEC958 device with no `plug` and may
also be silent; this PR only touches Standard.

## Tested configurations

Reproduced and fixed on two independent setups, both with main output mode
**Direct (`hw`)** and a USB DAC (DIYINHK), phone over A2DP (aptX-HD -> decoded
`PCM 16 bit 44.1 kHz`):

- **Raspberry Pi 3B, moOde 10.2.4 (stock release image), RPiOS 13 Trixie 64-bit**
  - Standard, no DSP/Peppy -> was silent (stock), now plays
    (`bluealsa-aplay ... -D plug:_audioout`, `_audioout` slave `hw:1,0`)
  - Compatibility -> unchanged (plays before and after)
- **Debian 13 x86_64, USB DAC**
  - Standard, no DSP/Peppy -> was silent, now plays
  - Standard + Invert Polarity -> was silent, now plays with phase inversion
  - Peppy / CamillaDSP / Graphic EQ / Crossfeed -> unchanged, still play + DSP/Peppy applies
  - MPD local playback (Direct/`hw`, bit-perfect) -> unaffected

The defect and the fix are pure ALSA routing and identical across platforms;
the code path does not branch on hardware.

## Files changed

- `www/inc/audio.php` — `updDspAndBtInConfs()`: Standard branch -> `plug:_audioout`
- `www/blu-config.php` — Bluetooth UI toggle handler -> `plug:_audioout`
- `www/inc/autocfg.php` — autoconfig reset -> `plug:_audioout`
- `etc/bluealsaaplay.conf` — default seed -> `AUDIODEV=plug:_audioout`

### DIFF
--- a/etc/bluealsaaplay.conf
+++ b/etc/bluealsaaplay.conf
@@ -1,5 +1,5 @@
 #########################################
 # This file is managed by moOde
 #########################################
-AUDIODEV=_audioout
+AUDIODEV=plug:_audioout
 BUFFERTIME=500000
--- a/www/audioinfo.php
+++ b/www/audioinfo.php
@@ -210,20 +210,7 @@ if ($_SESSION['invert_polarity'] == '1') {
 	$outputMode = $_SESSION['alsa_output_mode'];
 }
 
-// Bluetooth overrides
-if ($btActive === true) {
-	// Bluetooth inbound
-	if ($_SESSION['alsa_output_mode'] == 'iec958') {
-		$outputModeName = ALSA_OUTPUT_MODE_NAME[$_SESSION['alsa_output_mode']];
-	} else {
-		$outputModeName = ALSA_OUTPUT_MODE_BT_NAME[$_SESSION['alsa_output_mode_bt']];
-		$outputMode = $_SESSION['alsa_output_mode_bt'] == '_audioout' ?
-			$_SESSION['alsa_output_mode'] :
-			'plughw';
-	}
-} else {
-	$outputModeName = ALSA_OUTPUT_MODE_NAME[$_SESSION['alsa_output_mode']];
-}
+$outputModeName = ALSA_OUTPUT_MODE_NAME[$_SESSION['alsa_output_mode']];
 
 // Peppy ALSA
 $peppyAlsa = ($_SESSION['peppy_display'] == '1' || $_SESSION['enable_peppyalsa'] == '1') ? 'PeppyALSA &rarr; ' : '';
--- a/www/blu-config.php
+++ b/www/blu-config.php
@@ -117,19 +117,6 @@ if (isset($_POST['update_sbc_quality']) && $_POST['update_sbc_quality'] == '1') 
 	$_SESSION['notify']['title'] = NOTIFY_TITLE_INFO;
 	$_SESSION['notify']['msg'] = NOTIFY_MSG_BLUETOOTH_RECONNECT;
 }
-// ALSA output mode: Standard = _audioout, Compatibility = plughw
-// NOTE: conflicts with peppy
-if (isset($_POST['update_alsa_output_mode_bt']) && $_POST['update_alsa_output_mode_bt'] == '1') {
-	$_SESSION['alsa_output_mode_bt'] = $_POST['alsa_output_mode_bt'];
-	if ($_POST['alsa_output_mode_bt'] == 'plughw') {
-		$alsaDevice = $_SESSION['alsa_output_mode'] == 'iec958' ? getAlsaIEC958Device() : 'plughw' . ':' . $_SESSION['cardnum'] . ',0';
-	} else { // _audioout
-		$alsaDevice = $_POST['alsa_output_mode_bt'];
-	}
-	sysCmd("sed -i '/AUDIODEV/c\AUDIODEV=" . $alsaDevice . "' /etc/bluealsaaplay.conf");
-	$_SESSION['notify']['title'] = NOTIFY_TITLE_INFO;
-	$_SESSION['notify']['msg'] = NOTIFY_MSG_BLUETOOTH_RECONNECT;
-}
 // Controller mode
 if (isset($_POST['update_bluez_controller_mode']) && $_POST['update_bluez_controller_mode'] == '1') {
 	$_SESSION['bluez_controller_mode'] = $_POST['bluez_controller_mode'];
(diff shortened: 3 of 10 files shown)


moodeaudio:
With this update can the "ALSA output mode" option on Bluetooth Control screen be eliminated?

Gjuju:
Hi.
Honestly, I was wondering exactly the same thing about compatibility mode—and that whole option, really—but here’s what I think.

It could probably be removed safely, but I choose to keep it because Compatibility role has changed rather than disappeared.

Before this fix, Compatibility was the "at least it produces sound" escape hatch: on raw output modes (Direct/hw, IEC958) with no DSP/Peppy, Standard was silent, so Compatibility was the only working path. This PR removes that reason to exist, Standard now plays (theoricaly) in every configuration.

However, Compatibility and Standard are not functionally equivalent, they resolve to genuinely different chains:

Standard        -> AUDIODEV = plug:_audioout (added plug: to _audioout) 
bluealsa-aplay -> plug -> _audioout -> [DSP / Peppy / invpolarity] -> hw -> DAC

Compatibility   -> AUDIODEV = plughw:N,0
bluealsa-aplay -> plughw -> DAC   (bypasses the whole _audioout chain)

So Compatibility still does one thing Standard can't: it sends the Bluetooth input straight to the card, bypassing DSP, Peppy and polarity inversion, without having to disable those globally (which would also affect MPD). That makes it, post-fix, a "bypass processing / direct-to-card" option rather than a "make it work" option. It also doubles as a robustness/diagnostic fallback that doesn't depend on the shared _audioout definition.

In short, Compatibility is in fact a "Bypass DSP" mode now.
My concern is that someone with unfamiliar equipment might get stuck.

Let me know your thinking about this.
If any changes are needed, I'll make them without problem.

moodeaudio:
No worries, I'll go ahead and remove the option.

Gjuju:
omg...

i'm sorry i didn't saw you merged that PR.
 "ALSA output mode" option on Bluetooth Control screen is eliminated.

moodeaudio:
No problem. Thanks!

PR #759, opened 2026-07-03
Playlist view: default icon for playlists without a custom cover

Gjuju:
What — Playlists with no user cover (#EXTIMG:default) now show a default music-list icon in both the Playlist view tiles and the Edit-playlist cover preview, instead of the generic default-playlist-cover.jpg placeholder (+ text-name overlay).

before :
default_cover

after :
moded_cover

Why — putPlaylistCover() always materialises a cover file for every playlist — the user upload (local) or a copy of default-playlist-cover.jpg (default). So the onerror fallback added in #755 never fires for a default playlist: the placeholder loads fine (HTTP 200) and the user sees it instead of a clear "no cover" state. This keys the icon off the cover == 'default' metadata directly. The onerror fallbacks are kept for the residual case (an imported playlist whose referenced local cover file is genuinely missing).

Changes — command/playlist.php: get_playlist_contents now returns the already-computed cover field (was omitted), so the Edit modal can branch on it. js/playerlib.js: render the icon directly when cover == 'default'; factor the markup into PL_COVER_ICON/PL_PREVIEW_ICON reused by the fallbacks.

Testing — Validated on a moode-nopi box: a default playlist shows the icon in both views; local-cover playlists still show their image.

### DIFF
--- a/www/command/playlist.php
+++ b/www/command/playlist.php
@@ -297,7 +297,7 @@ switch ($_GET['cmd']) {
 		break;
 	case 'get_playlist_contents':
 		$playlist = getPlaylistContents($_POST['path']);
-		$array = array('name' => $playlist['name'], 'genre' => $playlist['genre'], 'items' => $playlist['items']);
+		$array = array('name' => $playlist['name'], 'genre' => $playlist['genre'], 'cover' => $playlist['cover'], 'items' => $playlist['items']);
 		echo json_encode($array);
 		break;
 	default:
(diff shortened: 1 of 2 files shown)


moodeaudio:
I noticed that when testing. There actually is an existing default cover` images/default-playlist-cover.jpg` that was never used, probably because I ran out of time back then.

In any case I went ahead and used it along with the new icon overlay instead of the fallback functions, added the css in panels.css, and added/modified various other bits and pieces in playlist.php, playerlib.js, indextpl.html

See commit
https://github.com/moode-player/moode/commit/b76491f760b69185e696783bf315429ee829c908

Gjuju:
Ok thanks for you reply.
I didn't seen that commit.
PR closed

PR #765, opened 2026-07-09
Reset the input-source scrim so static-message renderers render consistently

Gjuju:
Scope: this fixes an inconsistent-rendering bug, not the legibility of the
overlay text. See the note at the end.

The dark scrim (#inpsrc-style) is designed for the metadata screen: its
gradient is bottom-heavy (alpha 0.9 at 100%) and #inpsrc-metadata sits at
bottom:0. updateInpsrcMeta shows it, but nothing ever hides it again.

After a metadata renderer has been shown (AirPlay/Spotify/Deezer), the scrim
stays on and bleeds into the static-message screens drawn by inpSrcIndicator
(Bluetooth, Squeezelite, Plexamp, RoonBridge, Multiroom receiver), where the
message is centered and no scrim was ever intended. Two consequences:

- the same screen renders differently depending on which renderer ran before,
- the dark gradient ignores the Prefs > Theme setting on light themes.

inpSrcIndicator now hides the scrim, mirroring the updateInpsrcMeta call that
shows it. One line, symmetric with the existing show.

Test procedure
1. Set Prefs > Theme to a light theme, for example Whiteshade.
2. Reload the UI and connect Bluetooth first, without opening AirPlay/Spotify
   beforehand: the "Bluetooth Active" message shows with no dark scrim.
3. Start AirPlay or Spotify: the metadata screen shows the scrim behind the
   bottom text, as intended.
4. Return to Bluetooth: still no scrim; the screen is identical to step 2 and
   independent of navigation history. Before this change it kept the dark
   scrim left over from step 3.

Note on legibility
This PR does not make the overlay text legible when a cover backdrop is drawn
behind it (renderer_backdrop = Yes). That is a separate problem, and removing
the Cover Backdrop option is the right cure for it.

### DIFF
--- a/www/js/playerlib.js
+++ b/www/js/playerlib.js
@@ -803,6 +803,7 @@ function inpSrcIndicator(cmd, msgText) {
     $('#inpsrc-msg').css({width:'100%', top:'50%', bottom:'unset'});
     $('#inpsrc-metadata').hide();
 	$('#inpsrc-cover').html('');
+    $('#inpsrc-style').css('display', 'none');
 
     // Set the button and preamp volume
     // NOTE: Preamp volume #id will only exist if audioin != Local


moodeaudio:
You mean this issue, correct?
Its an old issue that arises because the backdrop image can't be reliably analyzed to determine whether the text and button colors should be light or dark.
Screenshot 2026-07-10 at 5 50 12 AM

Prolly the best way to fix it would be to eliminate the Cover Backdrop option altogether and revert back to the default backdrop scheme which does honor the Prefs > Theme setting. This should be no issue since AirPlay and Spotify cover art is now available.

Like below. 
Its using the modal backdrop color --modalbkdr: rgba(50,50,50,0.95); which gets tweaked a bit by the Theme engine but If you want less of the underlying WebUI showing the alpha blend could be bumped from 0.95 to 0.98.

Whiteshade Theme (Light)
Screenshot 2026-07-10 at 6 18 07 AM

Default Theme (Dark)
Screenshot 2026-07-10 at 6 18 47 AM

Gjuju:
Ok, honestly i thought it was a little bug because gradient/backdrop appear (in fact DO NOT disappear) after disconnecting spotify (or airplay) renderer. I looked at it a bit fast...

my workflow eg :
after reboot or F5 > connect BT renderer (no gradient/backdrop):
1-bt

then connect to spotify renderer :
2-spot

then back to BT renderer (with gradient/backdrop):
3-bt

I thought backdrop should be visible in BT family (btactive1, slactive1, paactive1, rbactive1, rxactive1, inpactive1), but in fact it must be force hidden, regardless of the precedent renderer right ?
I inverted the logic of that purpose, see las commit.


`Its an old issue that arises because the backdrop image can't be reliably analyzed to determine whether the text and button colors should be light or dark.
`

I will look at this then.

moodeaudio:
Adding image analysis back is a no-go with me.

There used to be image analysis plus auto-theme changing way back in the distant past but it was not accurate enough in picking the dominant color or dark/light levels in the image and users hated it so it was removed entirely. There was no renderer cover backdrop at that time.

Just try with Renderer Backdrop Off and test Light and Dark Themes in Prefs. The text and button colors should follow the Theme.

Gjuju:
Just a note on that PR.

This fix now ensure there is ALWAYS no visible backdrop BT family (btactive1, slactive1, paactive1, rbactive1, rxactive1, inpactive1) IF previous renderer used was (spotactive, aplactive, deezactive).

that's finally just a deterministic fix.

moodeaudio:
I don't understand this one.

Are you connecting via Bluetooth, then in your Spotify app selecting the Bluetooth connection or are you selecting the Spotify Connect connection?

Gjuju:
If you use spotify renderer THEN use bluetooth renderer you'll get the undesired backdrop.

with this fix there is no more undesired backdrop

so this case isn't possible anymore :
3-bt

moodeaudio:
Yes, if Spotify playback is stopped but not disconnected or client switched to Local playback, then connect via Bluetooth you end up with that messy visual. And even with the patch, I think the "Render Active" text and buttons would still be obscured depending on light/dark Theme and cover.

The challenge is that the renderer framework is not designed for renderers gracefully interrupting one another because most users only use a single renderer. It prolly could be coded to do interruption but I think only edge cases would benefit and thus you have to weigh effort/coding vs benefit. Also there is a concern over whether it would be reliable.

Gjuju:
You're right on both counts, and I should have been clearer about the scope —
the PR title was stale and overpromised. I've retitled it and rewritten the
description.

This PR does not fix legibility. Even with it applied, the "… Active" text and
buttons remain obscured whenever a cover backdrop is drawn behind them, exactly
as you say. Removing the Cover Backdrop option is the right cure for that, and
that's your call to make.

What it does fix is narrower: the scrim (#inpsrc-style) is shown by
updateInpsrcMeta and never hidden again, so it bleeds into the static-message
screens drawn by inpSrcIndicator, where a centered message was never meant to
have one. The result is that the same screen renders differently depending on
which renderer ran before it, and that a dark gradient is painted over light
themes, ignoring Prefs > Theme. The fix is a single line hiding the scrim in
inpSrcIndicator, symmetric with the existing show — no renderer interruption
logic, no new state, nothing to be unreliable. So I don't think the effort /
benefit and reliability concerns apply to this diff.

Worth noting: this reset is still needed after the Cover Backdrop option is
removed. Without it, the metadata scrim keeps leaking onto the centered-message
screens.

moodeaudio:
Ok I was able to repro the issue but its more involved then the whether the #inpsrc-style div is visible or not. 

In any case that div can't simply be hidden by the JS because its CSS rule defaults to display:none and thus it would always be hidden. Its purpose is to make sure the bottom buttons for AirPlay and Spotify display are on a dark background so they are visible thus its a top/bottom, light/dark gradient.

Whats actually happening is (1) the Spotify or AirPlay cover backdrop img .ss-backdrop is not being cleared when the renderer ends and (2) the #inpsrc-style div is not being reset to display:none

I might look into it tonight but we are under another Heat Dome > 90F so I might just chill and sip a Mojito.

moodeaudio:
I pushed a potential fix for this.
https://github.com/moode-player/moode/commit/4e1e5a35c5e7bafb0dc6c11a5edafbedab8d4d83

Gjuju:
perfect, it's globally the same fix.
thanks

PR #769, opened 2026-07-14
Peppy: bind the softvol control to the hardware mixer element (removes a double attenuation)

Gjuju:
With Peppy in the audio chain, a DAC that has a hardware volume control gets its volume attenuated twice — once in the DAC, once by a softvol control that should never have been created. This also makes the displayed dB describe only half of the real attenuation, which is enhancement item 10, Display accurate dB volume when Peppy is in the audio chain.

## Cause

`updPeppyConfs()` writes `$_SESSION['amixname']` into the softvol `control { name }` block of `peppy.conf`. That block takes a **raw ALSA control element** name, but `amixname` is the **simple mixer** name:

| | name |
|---|---|
| raw control element | `DIYINHK USB Audio 2.0  Playback Volume` |
| written by moOde | `DIYINHK USB Audio 2.0 ` |

The name never matches, so the hardware-control bypass in alsa-lib (`pcm_softvol.c`: `if (!(access & ACCESS_USER)) return 1;` → `*pcmp = slave`) never fires. Softvol creates its own user control instead of binding to the hardware one:

```
numid=4  'DIYINHK USB Audio 2.0  Playback Volume'  0-127  dBminmax -127.00 → 0.00     (hardware)
numid=9  'DIYINHK USB Audio 2.0 '                  0-255  dBscale  -51.00, 0.20/step  (softvol)
```

The simple-mixer layer then merges both elements under one name and writes the same percentage to both — measured on a DIYINHK USB DAC with Peppy on:

| MPD volume | softvol | hardware | real attenuation | dB displayed |
|---|---|---|---|---|
| 70 % | 236/255 → -3.8 dB | 118/127 → -9.0 dB | **-12.8 dB** | -8.97 dB |
| 40 % | 208/255 → -9.4 dB | 104/127 → -22.9 dB | **-32.3 dB** | -22.91 dB |

So the volume is attenuated twice, `getMappedDbVol()` reports only the hardware half, and software attenuation is applied to a DAC that has hardware volume.

## Fix

`getAlsaCtlElemName()` resolves the raw element name and `updPeppyConfs()` writes that, so softvol binds to the hardware control and bypasses itself. The peppyalsa meter and `_peppyout` stay in the chain (`*pcmp = slave`). Devices with no hardware volume are unchanged: `amixname == 'none'` → `PCM`, softvol legitimately created.

## Migration (second commit)

The stray control was persisted by `alsactl` in `asound.state` (`access 'read write user'`) and restored at boot, before MPD — so it kept shadowing the hardware control even with Peppy turned off. Existing installs would see no benefit from the fix alone, hence the purge in `worker.php`. `alsactl clean` only removes application-created controls, so a hardware element is never touched, and it is a no-op once clean.

## Result

After a reboot, playing, Peppy chain open:

```
peppy.conf  name "DIYINHK USB Audio 2.0  Playback Volume"
amixer      numid 2/3/4/5 only — no 0-255 control
sget        Limits 0 - 127 ; Playback 109 [86%] [-18.00dB]   (MPD volume 50)
```

The displayed dB now matches the real attenuation, applied once. PeppyMeter still runs.

One behaviour change: on a DAC with hardware volume the VU meter no longer follows the volume knob, since the digital stream is full-scale. That reactivity was a side effect of the double attenuation. Selecting *Software* MPD volume pins the DAC at 0 dB and attenuates upstream of peppyalsa, which keeps the meter reactive.

## Out of scope

In *Software* mode no dB is displayed at all, because `getMappedDbVol()` reads `amixer` and no ALSA control is driven. It could be computed from MPD's own curve (`gain = (exp(v / 25) - 1) / (e^4 - 1)`, `SoftwareMixerPlugin.cxx`), as the code already does for CamillaDSP. 
That is a separate enhancement, deliberately kept out of this fix.

### DIFF
--- a/etc/alsa/conf.d/peppy.conf.hide
+++ b/etc/alsa/conf.d/peppy.conf.hide
@@ -10,7 +10,7 @@ pcm.softvol_and_peppyalsa {
 type softvol
 slave.pcm "peppyalsa"
 control {
-# ALSA mixer (hardware mixer or "PCM" if none )
+# ALSA control element (hardware mixer element or "PCM" if none)
 name "PCM"
 card 0
 }
--- a/www/daemon/worker.php
+++ b/www/daemon/worker.php
@@ -745,6 +745,13 @@ workerLog('worker: ALSA mode:     ' . ALSA_OUTPUT_MODE_NAME[$_SESSION['alsa_outp
 // ALSA mixer
 phpSession('write', 'amixname', getAlsaMixerName($_SESSION['adevname']));
 workerLog('worker: ALSA mixer:    ' . ($_SESSION['amixname'] == 'none' ? 'none exists' : $_SESSION['amixname']));
+// Drop a stray softvol control left under the simple mixer name by an earlier release.
+// ALSA restores it at boot, where it shadows the hardware control. Only application
+// created controls are removed, so a hardware element is never touched.
+if ($_SESSION['amixname'] != 'none') {
+	sysCmd('alsactl clean ' . $_SESSION['cardnum'] . ' "name=\'' . $_SESSION['amixname'] . '\'"');
+	sysCmd('alsactl store ' . $_SESSION['cardnum']);
+}
 // HDMI mixer initialize (after first boot a test signal needs to be sent to "register" the mixer with ALSA)
 if ($_SESSION['alsa_output_mode'] == 'iec958') {
 	$result = getAlsaVolume($_SESSION['amixname']);
--- a/www/inc/alsa.php
+++ b/www/inc/alsa.php
@@ -68,6 +68,18 @@ function getAlsaMixerName($deviceName) {
 	return $mixerName;
 }
 
+// Get the raw ALSA control element name for a simple mixer control
+function getAlsaCtlElemName($mixerName, $cardNum) {
+	$elemName = $mixerName . ' Playback Volume';
+	foreach (sysCmd('amixer -c ' . $cardNum . ' controls') as $control) {
+		if (str_contains($control, "name='" . $elemName . "'")) {
+			return $elemName;
+		}
+	}
+
+	return $mixerName;
+}
+
 function getAlsaVolume($mixerName) {
 	$maxLoops = 3;
 	$sleepTime = 1;
--- a/www/inc/audio.php
+++ b/www/inc/audio.php
@@ -294,7 +294,11 @@ function updPeppyConfs($cardNum, $outputMode) {
 	}
 	sysCmd("sed -i 's/^slave.pcm.*/slave.pcm \"" . $alsaDevice . "\"/' " . ALSA_PLUGIN_PATH . '/_peppyout.conf');
 	// ALSA mixer
-	$alsaMixer = $_SESSION['amixname'] == 'none' ? 'PCM' : $_SESSION['amixname'];
+	// The softvol control{} block takes a raw control element name. A simple mixer name
+	// never matches, which makes softvol create its own 0-255 control alongside the
+	// hardware one instead of binding to it.
+	$alsaMixer = $_SESSION['amixname'] == 'none' ?
+		'PCM' : getAlsaCtlElemName($_SESSION['amixname'], $cardNum);
 	$peppyConfFile = file_exists(ALSA_PLUGIN_PATH . '/peppy.conf.hide') ? '/peppy.conf.hide' : '/peppy.conf';
 	sysCmd("sed -i 's/^name.*/name \"" . $alsaMixer . "\"/' " . ALSA_PLUGIN_PATH . $peppyConfFile);
 	sysCmd("sed -i 's/^card.*/card " . $cardNum . "/' " . ALSA_PLUGIN_PATH . $peppyConfFile);


moodeaudio:
it looks like it does fix the dB/attenuation issue but the side effect of no meter/volume tracking with hardware mixer is prolly going to be a no-go with users. I don't think they will like meters always at 0dB.

On the other hand meter tracking seems ok with MPD set to Software volume and forcing this is prolly an acceptable configuration when selecting Peppy but there are some other changes needed to support this.

For example the auto-switch from CamillaDSP volume back to default (hardware or software) doesn't check if Peppy is on and will always set MPD back to hardware if a mixer exists.

I'll test those additional changes a bit later today and post back.

Gjuju:
Agreed the meter behavior matters. You're right that with hardware volume the attenuation isn't reflected in the meter — the DAC attenuates downstream of Peppy's tap, so the needles read program level and don't follow the knob.

The reason it can't simply be made to follow the knob while keeping hardware volume: the softvol now binds to the hardware control, so alsa-lib drops it from the chain entirely (pcm_softvol.c: non-USER control → *pcmp = slave) and there's no software attenuation upstream of peppyalsa to meter. PeppyMeter can't recover the knob position either — its pipe data source ignores MPD's volume (set_volume() is dead code). So "hardware volume + knob-reactive meter" would mean patching PeppyMeter itself, not just our config.

But the knob-reactive meter already exists as an explicit user choice: picking Volume type = Software moves attenuation into MPD, upstream of peppyalsa, and the needles follow the knob. Since Hardware is only offered when the DAC has a hardware volume control, someone on Hardware chose it deliberately for bit-perfect — so I'd rather not auto-force Software when Peppy is on and override that choice. A help note explaining the trade-off is probably all that's warranted.

On the CamillaDSP point, yes, you're right, and i didn't saw this.
But it's broader than Peppy: that revert recomputes the mixer as "hardware if one exists, else software" ([worker.php:3160]), so it doesn't restore the user's actual prior Volume type — someone on Software who toggles CamillaDSP volume on/off lands back on Hardware.

On the Peppy angle specifically: the revert only needs to check Peppy if Peppy forces the mixer. In the approach above it doesn't — Peppy never touches Volume type — so there's no Peppy state to track. Restoring the user's saved choice is correct whether Peppy is on or off, and the meter behavior just follows from that choice (Hardware = program-level meter, Software = knob-reactive).

The primitive is already there: the Bluetooth path saves and restores the user's mixer via mpdmixer_local ([audio.php:160-166]). Having the CamillaDSP revert do the same instead of recomputing fixes it — a pre-existing rough edge, orthogonal to this PR; happy to take it separately.


EDIT //
There is already a PR that paves the way for volume attenuation to peppyMeter :
https://github.com/project-owner/PeppyMeter/pull/18

Gjuju:
I made a patch on PeppyMeter waiting for their PR#18 and a new waiting PR on moOde to fix definitely this problem and get volume tracking with hw volume too.

It's working like a charm with both hw and sw volume.

moodeaudio:
Tests complete.

This simple patch in worker always defaults to software volume from camilla volume if peppy is on.
```
Line 3107
//$mixerType = $_SESSION['alsavolume'] != 'none' ? 'hardware' : 'software';
if ($_SESSION['alsavolume'] == 'none') {
	$mixerType = 'software';
} else {
	if ($_SESSION['enable_peppyalsa'] == '1') {
		$mixerType = 'software';
	} else {
		$mixerType = 'hardware';
	}
}
```
There are a couple of other parts that would need to be patched for a complete solution but basically its just to (1) ensure that when peppy is turned on, MPD is set to software volume and (2) display dB for software volume (your other PR #770).

If I understand correctly, your concern would be that users might feel that software volume is inferior to hardware volume in some way. True there is some Audiophile psychology around this belief but volume attenuation is well established math that can be done either at the source or on the chip and it won't make any difference whatsoever in the quality of final bitstream that reaches the DAC unless the algorithm somehow has bugs. Volume attenuation also negates "bit-perfect" regardless of where in the pipeline it's done. If you want "bit-perfect" then volume has to be done externally, downstream from the DAC using an Analog volume control (which have their own issues).

Now for trade-offs I'd bet that users will readily accept software volume and meter tracking for Peppy rather than hardware volume and meters always at 0dB.

I did review PeppyMeter PR #18 and if I understand correctly it also needs your proposed enhancement to track live dB values (e.g. read from a small file) for hardware volume, right?

So prolly the best thing for me to do is merge and test PR #770 then wait for the outcome of the PeppyMeter PR and enhancements. If that works then go with it otherwise fallback to software volume for Peppy.

moodeaudio:
I was just thinking, if you don't want to wait for the peppy PR then you could do the following:

1. Make the necessary modifications to datasource.py to do volume tracking
2. Omit the new config var "volume.gain.db = 0" and configfileparser.py mods (Its not actually needed, correct??)
3. I'll add the modified datasource.py file to the package build for peppy-meter similar to how a modified peppyspectrum.py file was added to the peppy-spectrum package. See https://github.com/moode-player/pkgbuild/tree/main/packages/peppy-spectrum

Gjuju:
Yes, honestly to solve the entire problem we need to merge :
* this one #769 
+ #770 display db with sw volume
+ PeppyMeter PR https://github.com/project-owner/PeppyMeter/pull/18 
+ my new patch on PeppyMeter (waiting a few days for PR18  or i'll push my own) 
+ a new PR on moode to let it control the peppy attenuation...

then update peppy to the future "last version" if its all ok then we get :

| Combo | `mpdmixer` | Attenuation happens in | Vs peppyalsa tap | Meter follows the knob on its own | Correct |
|---|---|---|---|---|---|
| Peppy + **hw** | `hardware` | DAC | **downstream** | no | ✅ |
| Peppy + **sw** | `software` | MPD (top of chain) | upstream | yes | ✅ |
| CDSP + Peppy + **hw** | `hardware` | DAC | **downstream** | no | ✅ |
| CDSP + Peppy + **sw** | `software` | MPD | upstream | yes | ✅ |
| CDSP + Peppy + **CDSP vol** | `null` | CamillaDSP (playback device = `peppy`) | upstream | yes | ✅ |
| Peppy + **Fixed** | `none` | nothing | — | yes (full-scale) | ✅ |


You're right on my assumption on users and sw/hw volume.

moodeaudio:
I have 769 and 770 merged and they work great.

I'll wait to hear from you on your patch to peppymeter.

Gjuju:
> I was just thinking, if you don't want to wait for the peppy PR then you could do the following:
> 
> 1. Make the necessary modifications to datasource.py to do volume tracking
> 2. Omit the new config var "volume.gain.db = 0" and configfileparser.py mods (It's not actually needed, correct??)
> 3. I'll add the modified datasource.py file to the package build for peppy-meter similar to how a modified peppyspectrum.py file was added to the peppy-spectrum package. See https://github.com/moode-player/pkgbuild/tree/main/packages/peppy-spectrum

1. is required
2. that's not really the way i chose... here is what i understand :
volume.gain.db is a static dB number written directly in config.txt and read once at startup (changing it requires a restart, it's somehow a simple calibration var). 
I add :
volume.gain.db.source that is a path to a file that moOde (or any other player/bunch of code) rewrites on every volume change: PeppyMeter re-reads it each cycle, so the value is dynamic and follows the knob.
and configfileparser.py is modded too.

3. we'll see in a few days.

Gjuju:
I pushed a new PR #773.
It depends on PR https://github.com/moode-player/pkgbuild/pull/22 on pkgbuild

now you can test it.
best regards.

PR #770, opened 2026-07-14
Display the dB volume when MPD software volume is used

Gjuju:
Completes enhancement item 10 for the remaining case: with *Software* volume, no dB value was displayed at all.

## Why nothing was displayed

`getMappedDbVol()` reads the dB from `amixer`, but in software mode MPD attenuates the stream itself and drives no ALSA control — so the function returned an empty string and `Display dB volume` had no effect.

## What this does

The value is computed instead of read, exactly as the code already does for CamillaDSP (`CamillaDsp::calcMappedDbVol()`). MPD maps the volume with

```
gain = (exp(volume / 25) - 1) / (e^4 - 1)        PercentVolumeToSoftwareVolume(), SoftwareMixerPlugin.cxx
```

and `vol.sh` sends the knob level straight to `mpc volume`, so the displayed dB is exact:

| volume | 10 | 30 | 50 | 70 | 90 | 100 |
|---|---|---|---|---|---|---|
| dB | -40.7 | -27.3 | -18.5 | -10.8 | -3.6 | 0.0 |

DSD reports `0dB`, since MPD cannot attenuate it — the same caveat the Audio config page already warns about (*"DSD format will output 0dB when using Software volume"*). Reporting a computed value there would be misleading.

Hardware, Fixed and CamillaDSP volume are untouched.

## Testing

Verified on a USB DAC with the toggle on (`volume_db_display`): the displayed dB matches the table above at each knob position, and a DSD track shows `0dB` at any volume.

### DIFF
--- a/www/inc/mpd.php
+++ b/www/inc/mpd.php
@@ -724,7 +724,7 @@ function enhanceMetadata($current, $sock, $caller = '') {
 	if ($caller == 'engine_mpd_php') {
 		// Both these functions perform phpSession('open_ro');
 		$current['cover_art_hash'] = getCoverHash($current['file']);
-		$current['mapped_db_vol'] = getMappedDbVol();
+		$current['mapped_db_vol'] = getMappedDbVol($current['file']);
 		//debugLog('enhanceMetadata(): ' . $caller . ' OPEN_RO session');
 	}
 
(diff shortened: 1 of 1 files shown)


ppierre:
Thank you, it was upsetting me to not see the dB volume.

PR #772, opened 2026-07-15
Add "DAC quiet start" option to silence idle DAC noise at boot

Gjuju:
## Problem

Some DACs without an output mute relay emit a continuous hiss/crackle from
startup — typically a USB DAC right after it enumerates — and keep it until the
first track plays. moOde never feeds the device at boot: MPD restores
`state=stop` and, with `close_on_pause=yes`, never opens the output. So the DAC
sits making noise until playback starts. (The brief hardware pop at power-on is
separate and unavoidable.)

## Solution

A new **Configure → Audio → "DAC quiet start"** toggle (`cfg_system.dac_prime`,
default **off**). When enabled, moOde plays ~1 s of digital silence into the
selected output card the moment it initializes, which settles the DAC. It is not
USB-specific — an I2S DAC that hisses at power-on benefits the same way, and
priming any selected card is harmless (opt-in).

## How it works

- A udev rule fires on the sound-card `add` event, matching `KERNEL=="controlC*"`
  — the reliable sync point, since the ALSA control node is created last, after
  every pcm child (cf. `/lib/udev/rules.d/78-sound-card.rules`).
- The rule runs `systemd-run --no-block /usr/local/bin/moode-dac-prime %k`, which
  detaches the ~1 s `aplay` from the udev event (udev kills long-running `RUN`
  children). Same delegation pattern already used by `10-a2dp-autoconnect.rules`.
- The script primes **only the currently selected MPD output**
  (`cfg_mpd.device`), opening `plughw:<card>,0` directly since the DSP chain isn't
  up yet at enumeration time.
- Being tied to the `add` event, it works at **boot and on hot-plug**, no polling.

## Enable / disable

The rule ships as a versioned file, **disabled**
(`89-moode-dac-prime.rules.disabled`). `worker.php`'s `applyDacPrime()` is the
single authority: it enables/disables the rule by renaming `.disabled ⇄ .rules`
according to `dac_prime`, both at worker startup and on the toggle's job. Since
the rule is a persistent file, once enabled it triggers at enumeration on later
boots without waiting for the worker.

## Files

- `var/local/www/db/moode-sqlite3.db.sql` — new `dac_prime` param (default off)
- `www/snd-config.php`, `www/templates/snd-config.html` — the toggle (mirrors the
  existing Autoplay control)
- `www/daemon/worker.php` — `applyDacPrime()`, boot-time application, and the
  `dac_prime` job
- `etc/udev/rules.d/89-moode-dac-prime.rules.disabled` — the udev rule, shipped
  disabled
- `usr/local/bin/moode-dac-prime` — the prime script (packaged executable, like
  `a2dp-autoconnect`)

## Notes

- Default **off** — opt-in, no behaviour change for existing setups.
- The enumeration pop itself is hardware and unaffected; this only clears the
  residual idle noise that follows it.
- Tested on Raspberry Pi. There's a short delay (~1–2 s) between enumeration and
  the silence taking effect, due to the `systemd-run` transient-unit spin-up.


moodeaudio:
I'm not opposed to adding certain workarounds for badly behaving DAC's but is there a particular brand DAC in your system thats got this noise issue at startup?

Gjuju:
Yeah, that's my main DAC and it's DIY :  
https://www.diyinhk.com/shop/eol-product/68-768khz32bit-ak4490eq-dac-i2sdsd-input.html

Love it but it set itself in DSD mode just after usb enumeration an sometimes give some low crackle sounds, sometime give some kind of huge white noise and the problem remains unless it plays PCM.

Ready Script / ReadyChime partialy do the job but it remains 20sec of potential oddity before it plays. 

I found an old thread where it could be a benefit, but I aggree it seems to be really rare case.  
https://moodeaudio.org/forum/showthread.php?tid=733&highlight=usb+AND+noise

PR #773, opened 2026-07-15
Peppy: make the VU meter follow a hardware-volume knob

Gjuju:
The `peppyalsa` scope taps before the DAC, so hardware volume — attenuated *in the DAC*, downstream of the tap — doesn't move the needles. moOde now publishes the current attenuation dB to `/tmp/peppy_gain_db`, and PeppyMeter scales its display through `volume.gain.db.source`. Hardware mixer only; software / CamillaDSP / Fixed attenuate upstream of the tap already, so `0` (unity) is written and nothing changes there. 100% → 0 dB → non-regression.

~~Build addressing @moodeaudio's meter-UX concern on #769 (needles stuck at full scale with hardware volume).~~ / **Merged**
~~**Depends on** moode-player/pkgbuild#22. Without it the config key is ignored (graceful fallback).~~ / **Merged**.

## Design: one writer, on the ALSA control

Every volume path — the WebUI, `vol.sh` (USB knob, IR remote), `rotvol.sh`, `mpc`, an external `amixer` — ends up moving the same ALSA control. A small daemon watches that control (`alsactl monitor`) and republishes the dB on every change, so one writer covers them all and no future volume path has to remember to publish. `alsactl` comes from `alsa-utils`, already a dependency — no new one.

- `peppy-gain.php` publishes once at startup (`/tmp` is empty at boot, and PeppyMeter falls back to unity while the file is missing), then on each event. It is **sessionless**: it runs as root via `sysCmd()`, and opening the PHP session would recreate the session file root-owned. It reads `cfg_system` directly, like `touchmon.php`.
- `getMappedDbVol()` is untouched — the ALSA read is split into `getAlsaMappedDbVol()`, taking args so the daemon can call it without a session.
- Lifecycle is worker-owned, like the LCD/GPIO daemons: started at worker startup, bounced by `updPeppyConfs()` when the card (`hw:N`), output mode or volume type changes, and restarted from the polling loop if it ever dies — otherwise the Peppy gain would freeze and the needles would silently stop following (audio is unaffected either way).
- Gated on `peppy_display || enable_peppyalsa`, the same test `updAudioOutAndBtOutConfs()` uses to route `_audioout` through peppyalsa. `peppy_display` alone is not enough: touchmon writes it straight to the database to say which screen is showing *right now* (it swaps to the WebUI whenever playback stops), so a worker started with the WebUI on screen would never run the monitor, and playback would then bring Peppy back with no gain file at all.

## Tested on hardware (Pi 3B+ and x86 both 10.3.0, USB DAC with hardware volume)

- `/tmp/peppy_gain_db` tracks the DAC dB curve (100%→`0`, 50%→`-18`, 0%→`-127`), matching the dB the UI displays — one source of truth.
- Needles follow the **WebUI knob** and a **USB keyboard through triggerhappy** (`vol.sh -up/-dn`), the latter with no browser involved at all.
- Software and CamillaDSP volume: `0` written, meter unchanged.
- Reboot-validated: the gain is published before playback starts, so the needles are right on the first frame.
- Killed the daemon with `kill -9`: back within two worker ticks, its stray `alsactl` reaped by the restart, and the worker logs it.

*Force-pushed twice: the design moved from a hook in `enhanceMetadata()` (which only tracked the volume while a WebUI client was polling) to the event-driven writer above, after testing showed a knob or remote left the needles on a stale gain.*


moodeaudio:
Looking good. I wont be able to test until I finish the Radio Browser fixes and updates.

Gjuju:
Thanks.

Hope you seen the last force push.

Take you time on RB, if you need some help, you're welcome.

moodeaudio:
Haven't looked at it yet. 

RB doesn't need much work, just some updates and fixes. The majority of the contrib is excellent :-)

Gjuju:
@moodeaudio 
PeppyMeter merged the new volume.db.source file feature on his master.

It depends on a new PeppyMetter:master install.
Best regards  

EDIT //
On older install /etc/peppymeter/config.txt need to be updated to exposes the 2 new [data.source] keys, I'm not sure on the best way to do it.

Do you provide an update script on new releases ?

moodeaudio:
I'm still debating what the best solution is for the Hardware volume issue. See the latest update to the Challenge list:  10. [WIP] Display accurate dB volume when Peppy is in the audio chain. https://moodeaudio.org/forum/showthread.php?tid=8474

Since this PR requires the most recent PeppyMeter that includes PR #20 from that repo I'd need to build a new package and propagate that to the image builder and in-place updater. I'll look into that today.

moodeaudio:
Give the new package a try
```
sudo apt update
sudo apt -y install peppy-meter=2026.7.20-1moode1
```

Update the config
```
sudo sed -i -e 's/^framebuffer.device.*/framebuffer.device = \/dev\/fb0/' \
-e 's/^mouse.device.*/mouse.device = \/dev\/input\/event0/' \
-e 's/^mouse.enabled.*/mouse.enabled = False/' \
-e 's/^pipe.name.*/pipe.name = \/tmp\/peppymeter/' \
/etc/peppymeter/config.txt
```

Gjuju:
Are you sure of the update config ?

it shoul only add 2 lines *volume.gain.db = 0* and *volume.gain.db.source =*

`sudo sed -i '/^volume.max.in.pipe = 100\.0$/a\
volume.gain.db = 0\
volume.gain.db.source =
' /etc/peppymeter/config.txt`

moodeaudio:
The full config update is needed after the standard package install `sudo apt -y install peppy-meter=2026.7.20-1moode1` because it overwrites config.txt. This is whats done as part of an image build.

For an in-place update scenario you are correct that only the two lines need to be added but the apt command would be different so it leaves the existing conf.txt as-is.

Try this
```
sudo apt -y -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install peppy-meter=2026.7.20-1moode1
```

Gjuju:
It's just perfect !
 
1 - Installed peppy-meter=2026.7.20-1moode1 + your sed :
peppyMeter is working as before ---> HW volume show full volume **without** attenuation
**No regressions**

2 - added :
volume.gain.db = 0
volume.gain.db.source = /tmp/peppy_gain_db
peppyMeter is working as before ---> HW volume show full volume **without** attenuation
**No regressions**

3 - gulp deploy this contrib and all is working as inteded
peppyMeter is working as before ---> HW volume show volume **with** attenuation
**Works Ok**


EDIT // just edited the full message.

moodeaudio:
Ok, good.

There is one usage scenario the simpler "Software volume" approach won't handle and thats playback of DSD files. They are always output at 0dB by MPD (native or on-the-fly PCM).

I'll review and test this PR a bit later.

moodeaudio:
Can you remove the config.txt file? It's causing a PR conflict with the one I committed earlier as part of the bump to peppymeter.

Thx

Gjuju:
Remove or resolve conflict is Ok ?

I'll remove in 1h30 is necessary.

EDIT // 

- DSD isn't unsuported by PeppyMeter and will crash it
- DoP trasport can't be read/passthrough by PeppyMeter's FIFO, but DoP amplitude carry only a constant flux of 0400 0400 0400… --> peppy will display but volume remains 0 and needles does'nt move.

I didn't try before as I only have 1 DSD album.

PeppyMetter probably never worked whith DSD or DoP, am I right ?

moodeaudio:
Works great :-)
Needs a few very minor comment updates.

Yes, native DSD and DoP not supported by peppy, it only supports PCM. The use case I was referring to earlier is MPD doing on-the-fly DSD to PCM when DAC indicates no support for DSD native. The converted PCM is at 0dB and thus the Software volume approach is not a universal solution.

Issue #23, opened 2019-03-26
SyntaxError: At least one digit must occur after a decimal point


piotr277:
With Moode 4.4 the player UI is not displayed correctly with my Blackberry Playbook. No reactions on clicking, only blank page is displayed:
![Zdjęcie, 26 03 2019 o 08 01](https://user-images.githubusercontent.com/18440238/54977521-da279e00-4f9d-11e9-80a4-39d7b4fffd64.jpg)
After some digging in I've found that there is problem in playerlib.js which may cause these problems:
<img width="724" alt="Zrzut ekranu 2019-03-26 o 07 59 47" src="https://user-images.githubusercontent.com/18440238/54977312-30481180-4f9d-11e9-9718-3cad7be2e4c1.png">

pccasto:
Did you use your Blackberry Playbook to connect to a previous version of moOde? What you are seeing (visually) looks a lot like the symptoms of having cached css and js files from a previous version.

Can you clear the cache and try again? I would expect the syntax/reference error to clear.  The bgimage not found might continue, but that is harmless.

The problem of stale cache files on upgrade has been fixed in 5.0beta.

~paul

piotr277:
> Did you use your Blackberry Playbook to connect to a previous version of moOde? 

No, I didn't. I did the upgrade from 4.3 to 4.4 before I used Playbook.
I've also cleared the cache but it doesn't help.

The `SyntaxError: At least one digit must occur after a decimal point` happens in `playerlib.js` in this line: `const FEAT_ADVKERNELS =  0b0000000000000001;	//     1`

monkeypox:
It's weird that line would be a problem since earlier moodes also used the same construct, your console might be ignoring comment lines, if you don't have a way to inspect it try adding something like...

var test123 = '';

...right after that line, if it really is the problem then the error line number will stay the same otherwise it'll shift by 1. If it does shift you can use the same method to find the actual line by moving you put the var test line down the source and stopping when the line in the console doesn't shift.

pccasto:
What version of JavaScript does your playbook browser use?
A literal "0b[01]+" is valid in ECMAScript 6, and is interpreted as a binary value
So ```const FEAT_ADVKERNELS = 0b0000000000000001;``` would be valid with that version.

I'd expect your version might predate ECMAScript 6 (pre 2015), so may not handle that syntax.
That would explain why it is throwing an error message when it tries to parse that line.
[...]

PR #779, opened 2026-07-22
Peppy: guard the meter against native DSD (temporary, until alsa-lib is fixed)

Gjuju:
Since #769 the Peppy chain is transparent on hardware-volume DACs — which is correct, but exposes a latent alsa-lib defect: type meter's s16 scope has no DSD conversion and aborts the caller (pcm_meter.c:1222), killing MPD on Peppy + DSD-capable DAC + hardware volume + dop "no" + a DSD file. Before #769 the stray softvol refused DSD by accident, so MPD always converted and the bug never fired.

This adds a unity route above the meter that restores exactly that behaviour, deliberately: native DSD is refused at negotiation, MPD falls back to dsd2pcm, the meter reads PCM. Unity gain, no control created (no double attenuation), bit-transparent for PCM, and DoP passes unharmed (markers verified intact below the route). Transparent to users — nothing to enable.

Validated on a fresh Pi 3B (stock 10.3.1 develop, stock alsa-lib/peppy-alsa): the crash config plays S32_LE @352800 with live needles, 0 aborts; DoP unaffected.

Temporary by design: revert when a meter-fixed alsa-lib ships (moode-player/pkgbuild#24, or upstream alsa-project/alsa-lib#516/#517 reaching a base OS). Existing installs would need the postinstall to lay the updated file.

### DIFF
--- a/etc/alsa/conf.d/peppy.conf.hide
+++ b/etc/alsa/conf.d/peppy.conf.hide
@@ -3,7 +3,23 @@
 #########################################
 pcm.peppy {
 type plug
-slave.pcm "softvol_and_peppyalsa"
+slave.pcm "peppy_pcmguard"
+}
+
+# Only negotiates linear PCM: native DSD is refused here so the player
+# converts upstream instead of reaching the meter, whose s16 scope
+# aborts on DSD. Unity gain, bit-transparent. Remove when alsa-lib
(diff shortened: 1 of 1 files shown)


Gjuju:
@moodeaudio 
Seems useless untill you merged the pkgbuild #23/24/25

I close.

PR #780, opened 2026-07-24
Bluetooth: don't store the "Pincode set" placeholder as the PIN code

Gjuju:
## Problem

With a Bluetooth PIN code set, the input field shows the `Pincode set` placeholder.
Submitting the Renderers page again stores that string as the PIN: `/etc/bluetooth/pin.conf`
ends up holding `* Pincode set` and the code the user chose is lost.

The guard meant to prevent this tests the submit button's value instead of the input
field, so it never matches.

## Fix

- `www/ren-config.php` — test the input field, like the Deezer, Qobuz and SMB pages
  already do for their own `Password set` placeholder

## Testing

Pi 3B: reproduced before the change, PIN preserved after, and clearing the PIN still works.


Gjuju:
Is now useselss with BT refacto on #783 .

PR #781, opened 2026-07-24
Bluetooth: don't let a failed hciconfig call kill the pairing agent

Gjuju:
## Problem

Setting a Bluetooth PIN code adds `hciconfig hci0 sspmode 0` to `bt-agent.service`.
Some controllers reject that call — a USB dongle here returns
`Can't set Simple Pairing mode on hci0: Input/output error (5)`. Without a `-` prefix
systemd fails the whole unit, so `bt-agent` is stopped and Bluetooth pairing no longer
works at all until the PIN code is removed.

## Fix

- `etc/systemd/system/bt-agent.service` — prefix the call with `-` so a rejected
  `sspmode` no longer fails the unit
- `www/daemon/worker.php` — widen the two `sed` patterns so they keep matching the line
  once it carries the prefix, otherwise the `sspmode` toggle stops working silently

## Testing

Pi 3B: the call succeeds, nothing changes and the toggle still works. Machine with a USB
dongle that rejects it: `bt-agent` goes from `failed` to `active`.

### DIFF
--- a/etc/systemd/system/bt-agent.service
+++ b/etc/systemd/system/bt-agent.service
@@ -8,7 +8,7 @@ Type=simple
 KillSignal=SIGINT
 ExecStart=/usr/bin/bt-agent -c NoInputNoOutput
 ExecStartPost=/bin/sleep 1
-ExecStartPost=/bin/hciconfig hci0 sspmode 1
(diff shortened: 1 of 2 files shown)


Gjuju:
Is now useselss with BT refacto on #783 .

PR #783, opened 2026-07-25
Bluetooth: pairing confirmation instead of the PIN code

Gjuju:
## Problem

The Bluetooth PIN code option no longer works with modern devices. Phones and
computers now require an authenticated (man-in-the-middle protected) pairing,
and a fixed PIN cannot provide that — the pairing is rejected before it
completes. So the PIN field is present in the UI but never actually pairs
anything.

## What this does

It replaces the PIN with a **pairing confirmation** flow, the same one used by
commercial Bluetooth speakers with a screen.

When a device tries to pair, moOde and the device each show the **same 6-digit
code**. A dialog pops up in the WebUI — and on the local display when it is
enabled — showing the device name and the code, with **Confirm** and **Reject**.
The user checks the code matches what the phone shows and confirms. This produces
an authenticated, secure pairing that today's devices accept.

The confirmation is driven by the Bluetooth stack, so there is nothing to type
and nothing to remember, and the dialog closes on its own if the device gives up.

In Configure → Renderers, the PIN code field becomes a **Pairing confirmation**
On/Off toggle:

- **On** (default): the confirmation dialog described above.
- **Off**: devices pair directly with no confirmation (Just Works), as before.

The old PIN plumbing is removed.

## Testing

Tested end to end on real hardware: a phone that refuses the old PIN pairs
successfully through the confirmation dialog, the pairing is authenticated, and
audio plays. Toggling it Off pairs without a dialog. Rejecting, or letting the
device time out, cancels cleanly.


moodeaudio:
Are you sure the PR is ok? There are tons of files from your nice "nopi" project.

Gjuju:
yes you're right.
that was a misclick !

see #784 

I close this one

PR #788, opened 2026-08-13
Show the actual Triggerhappy state instead of defaulting to off

Gjuju:
usb_volknob is a session-only , but the state that actually persists across reboots is triggerhappy's service enable state — the usb_volknob job does systemctl enable/disable. The session var is only a mirror of it.

When the session var is missing, worker.php assumed off and wrote that default.
After any session reset the UI then reports the USB volume knob as off while thd is still enabled and running, and the toggle can no longer turn it off: per-config.php submits the job only when the posted value differs from the session one, so clicking OFF on an already-0 session var is a no-op — you have to go ON then OFF to regain control.

This refreshes the session var from systemctl is-enabled triggerhappy at every worker start, so the UI and the worker: log line show the running state. 

Note: led_state has no equivalent fix — nothing persists it outside the session file, and at boot /sys/class/leds still holds the firmware default, so reading it back would yield the same 1,1 as today.


moodeaudio:
What would cause the session var to be deleted after its initially set to the default of 0?

Gjuju:
It's a marginal case.

Any Pi user restoring a backup hits this: backup_manager.py extracts moodecfg.ini to /boot, autocfg then sets usb_volknob via setSessVarOnly, so the UI comes back showing On while triggerhappy was never really re-enabled.

moodeaudio:
Looking at the code in worker I see that the problem seems to be a missing conditional to start/stop Triggerhappy during worker startup. This guarantees the service state will reflect the session var value after a restore.

Exiting code
```
// Log Triggerhappy / USB volume knob on/off state
if (!isset($_SESSION['usb_volknob'])) {
	$_SESSION['usb_volknob'] = '0';
}
workerLog('worker: Triggerhappy:     ' . ($_SESSION['usb_volknob'] == '1' ? 'on' : 'off'));
```

Should be:
```
// Start Triggerhappy / USB volume knob
if (!isset($_SESSION['usb_volknob'])) {
	$_SESSION['usb_volknob'] = '0';
}
if ($_SESSION['usb_volknob'] == '1') {
	sysCmd('systemctl enable triggerhappy');
	sysCmd('systemctl start triggerhappy');
} else {
	sysCmd('systemctl stop triggerhappy');
	sysCmd('systemctl disable triggerhappy');
}
```

I don't think there is a need to consider the case where the session variable 'usb_volknob' or any session variable for that matter could somehow end up deleted from the session file since all the session variables are established after the first-boot of a moode system and they persist forever.

I'll go ahead and push the change described above. We can revisit if needed.

Gjuju:
Understood

PR #790, opened 2026-08-20
Qobuz Connect renderer

PhilipVinc:
Hello, 

With the help of Claude I spent the last 2 days trying to get qbzd working as a renderer inside of Moode.
To make it work in the same way as Spotify connect (no login on the moode side), and to fix several issues in how qbzd reports it's state (Buffering, loading, etc) I had to do a lot of changes to qbzd. 

I will eventually try to upstream those, but as that would require a bit more time which I won't have over the next few weeks, so it's more of a longer term plan. 
Moreover, there's still a few issues I'd like to address, so the code for now lives at https://github.com/PhilipVinc/qbz/tree/feature/external/qbzd-connect-pairing and I'm tagging releases at https://github.com/PhilipVinc/qbz/releases .

This PR is the moode side of things. Also claude-generated largely, with a lot of steering and manual testing to identify issues. 
The implementation basically follows the same hook system as the other renderers. There's nothing fancy there (note that I did not try the multi receiver implementation).

Would you be willing to accept this PR, assuming I clean it up to standards, and assuming that I will be venturing my own fork of qbzd until I manage to upstream the changes, or that would not be acceptable to you?

### DIFF
--- a/etc/alsa/conf.d/qbzd-devices.conf
+++ b/etc/alsa/conf.d/qbzd-devices.conf
@@ -1,0 +1,9 @@
+#########################################
+# This file is managed by moOde
+#########################################
+# ALSA hints so the qbzd Qobuz Connect renderer can select
+# moOde's output chains by name
+pcm._audioout.hint.show on
+pcm._audioout.hint.description "Moode Audio Output"
+pcm.btstream.hint.show on
+pcm.btstream.hint.description "Moode Bluetooth Stream"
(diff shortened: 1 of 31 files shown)


PhilipVinc:
Superseded by #799.

The renderer itself is on `develop` now, so this draft's 38 commits are behind it — they still pin `qbzd 2.0.2.moodeNN`, and the daemon has since been renamed to Pibuz. #799 is the follow-on against current `develop`: it points the renderer at `pibuz`, and fixes three settings on the Qobuz Connect config screen that turned out to do nothing.

PR #793, opened 2026-09-09
Reset the Qobuz Connect active flag on worker startup

Gjuju:
`chkRendererActive()` clears every renderer's active flag at startup, but the
Qobuz one is spelled `qbzctive` there while the parameter is `qbzactive`, so the
flag is never cleared and the renderer screen can come up already "active".

This PR carried three fixes; the pibuz rename landed the other two --
`isQobuzUpgradable()` queries pibuz now, and the `quality_fallback` setting the
third one touched has been removed. Rebased on develop, this is what's left.


PhilipVinc (approved):

Gjuju:
Refactored in #795

PR #795, opened 2026-09-10
Qobuz Connect: two worker fixes

Gjuju:
Two independent fixes on the worker's Qobuz path.

**Saving settings no longer ends the Connect session.** A save ran
`stopQobuz(); startQobuz();`, which takes the daemon down: playback stops and the
app has to be pointed at the player again. pibuz applies a settings change to a
running daemon, so a save now just sends it the settings.

Two settings are the exception: `alsa_buffer_ms` and `memory_cache_mb` are read
only when the daemon starts, so changing either still restarts it — and the
notification says so instead of claiming the change was applied.

**The Qobuz active flag is now cleared at startup.** `chkRendererActive()` spells
it `qbzctive` while the parameter is `qbzactive`, so it was never reset and the
renderer screen could come up already "active".

`startQobuz()` is split into `cfgQobuz()` + `startQobuz()`, the shape
`cfgSqueezelite()`/`startSqueezelite()` already uses. The service toggle and the
manual restart button are unchanged.

Tested on x86_64 with pibuz 2.4.2-dev.1: saving Quality during playback leaves
the daemon's pid untouched and the music running; saving ALSA buffer restarts it,
as intended.


PhilipVinc:
i think thats a good improvement!

PR #796, opened 2026-09-11
Enable the Qobuz Connect install button

Gjuju:
The block was commented out because the recipe and the plugin zip did not exist. 
Both are proposed now: pkgbuild#29 and plugins#1. 
This removes the temporary notice and the comment markers, nothing else.

Note: isQobuzUpgradable() queries `dpkg --show librespot` instead of qbzd (renderer.php:360, third fix in #793), so the button reads "Upgrade" even when qbzd is up to date. Verified on an Orange Pi 3 LTS.


Gjuju:
Just edited to aligns cfg_plugin.version with the package the recipe now builds (qbzd_2.0.2-49-1moode1), since the plugin derives the .deb file name from it.

With pkgbuild#29 merged and plugins#1, the chain then runs end to end.

Gjuju:
Closing -- develop does this now. The install button is enabled upstream, with
the isPibuzInstalled() / official-method split, and cfg_plugin already reads
v2-pibuz 2.4.0-1moode1. Nothing left in this branch.

PR #799, opened 2026-09-14
Qobuz Connect: point the renderer at Pibuz, and fix three settings that did nothing

PhilipVinc:
Ready to go! 

This updates the Qobuz connect renderer with the rename of qbz to Pibuz 2.4.0 and, most importantly, by cleaning up some unused options that were not exposed anymore, and by cleaning up the presentation of other options.

Fore example, 
- gapless playback is now always available.
- On Pi3B+ or newer, one can decide to turn off the disk cache, or size it accordingly (this allows faster seek/previous/next song)
- other things, see below:

The settings table changed because I had a lot of cruft/old settings leftover from the development. 

## Notes for review

- `cfg_qobuz` gains three params, taking one previously unused row and reusing
  two that nothing read. `qbz-config.php` now inserts any missing param with its
  default before rendering, so a player upgraded in place does not show a blank
  control — the standalone package already did this and mainline did not, which
  is how one of the dead settings above stayed invisible.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


PhilipVinc:
@moodeaudio this is good to go! I think I'm done with this stuff.

It's changing the settings table a bit because there were a few dead features, and a few new ones...
I've also inserted some details on how to edit some more 'advanced' settings that might be helpful for very advanced users.

I just have one question: instead of requiring building on the pi, can't you make it download the binary directly? I build it on GitHub actions so it is available, and it's built against an old toolchain so to be compatible with any reasonable LIBC version...

Or you have some argument for why you prefer to force compilation on device?

PR #801, opened 2026-09-17
rotenc.py: decode quadrature with a state table to stop spurious reverse steps

nitrousptm:
The ISR derived the rotation direction solely from which pin fired the edge that returned the encoder to the idle 1-1 state. That breaks in two ways with a mechanical encoder:

- Contact bounce on the first contact to close (e.g. A when turning CW) briefly returns the inputs to 1-1 via pin A, which is counted as a CCW step before the real CW step follows. If the two land in different poll intervals, the volume visibly jumps back.
- RPi.GPIO delivers callbacks late and serialised. When both rising edges arrive within that latency (fast rotation or a busy CPU), the callback for the first edge already reads 1-1 and attributes the step to the wrong pin.

Replace the pin-based heuristic with Ben Buxton's full-step quadrature state table. A step is emitted only after the complete 11-01-00-10-11 (or reverse) sequence, independent of which pin raised the interrupt. Bounce or a merged edge can at worst drop a step; it can no longer produce one in the wrong direction. No hardware debounce is required.

Tested on a Pi 4B with a Bourns PEC11R-4220F-S0024 (24 detents, 24 PPR) on GPIO 23/24 with internal pull-ups: 10 detents CW/CCW consistently yield 10 steps in one direction, both slow and fast, where the previous code produced intermittent reverse steps.

Compatibility: direction mapping is preserved (A closing first = CW = volume up), so existing installations keep their rotation sense. CLI arguments and the accel/step handling in the poll loop are unchanged. Encoders with 2 or 4 detents per electrical cycle still emit one step per full cycle, exactly as before.


moodeaudio:
Looks good.

